PR #6, opened 2021-08-04
Resolved FastifyDeprecation warning and added typescript documentation.

vikasg603:
This pull request does
1) Resolve FastifyDeprecation warning, used https://github.com/SerayaEryn/fastify-session/pull/143
2) Added typescript documentation about the custom sessions.
3) Removed deprecated FastifyPlugin import and instead added FastifyPluginCallback.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ app.addHook('preHandler', (request, reply, next) => {
   next();
 })
 ```
-**NOTE**: For all unencrypted (HTTP) connections, you need to set the `secure` cookie option to `false`. Look below for all cookie options and their details.  
+**NOTE**: For all unencrypted (HTTP) connections, you need to set the `secure` cookie option to `false`. See below for all cookie options and their details.  
 The `sessionStore` decorator of the `request` allows to get, save and delete sessions.
 ```js
 app.register(fastifySession, {secret: 'a secret with minimum length of 32 characters'});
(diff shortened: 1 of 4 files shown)


lib/fastifySession.js
```diff
@@ -18,8 +18,8 @@ function session (fastify, options, next) {
   fastify.decorate('decryptSession', (sessionId, request, callback) => {
     decryptSession(sessionId, options, request, callback)
   })
-  fastify.decorateRequest('sessionStore', options.store)
-  fastify.decorateRequest('session', {})
+  fastify.decorateRequest('sessionStore', { getter: () => options.store })
+  fastify.decorateRequest('session', { getter: () => ({}) })
```
climba03003, line 22:
```suggestion
  fastify.decorateRequest('session', { getter: () => ({}) })
```
You cannot change it with `getter` only. Please change it to `null` and add a empty object before or on `preValidation` hook.

vikasg603: Easy fix.
My bad.

README.md
```diff
@@ -119,6 +119,20 @@ fastify.decryptSession(sessionId, request, () => {
 })
 ```
 
+### Typescript support:
+This plugin support typescript, and you can extend fastify module to add your custom session type.
```
zekth, line 123:
```suggestion
This plugin supports typescript, and you can extend fastify module to add your custom session type.
```

zekth (changes requested):
minor typo

vikasg603 (commented):
Final review

lib/fastifySession.js
```diff
@@ -18,8 +18,8 @@ function session (fastify, options, next) {
   fastify.decorate('decryptSession', (sessionId, request, callback) => {
     decryptSession(sessionId, options, request, callback)
   })
-  fastify.decorateRequest('sessionStore', options.store)
-  fastify.decorateRequest('session', {})
+  fastify.decorateRequest('sessionStore', { getter: () => options.store })
```
climba03003, line 21: I am not sure about it. The `store` can be changed before but it cannot change now.

If it is `getter` only, the type should be changed to `readonly`.

vikasg603: Hey @zekth, any comment on this ?

zekth: Indeed this is a readonly value, please adjust the typings

vikasg603: Hey, @zekth, I am not sure about what you are talking about?
Will you suggest to me the changes that need to do?

mcollina (approved):
lgtm, this is good

lib/fastifySession.js
climba03003: @vikasg603
Change it to readonly. https://github.com/fastify/session/blob/master/types/types.d.ts#L10
Add test-case inside https://github.com/fastify/session/blob/master/types/test/using-store.ts

It should be something like this, we expect it to be throwing error
```ts
// @ts-expect-error
request.sessionStore = null
```

climba03003 (approved):
LGTM.

zekth (approved):
lgtm! thanks

PR #14, opened 2021-08-13
fix(types): disallow unknown properties on session

SimenB:
Port of https://github.com/SerayaEryn/fastify-session/pull/145, original OP below.

---

Picks up typos like `session.sessionID` (which is what it's called in `express-session`).

Possibly breaking change, but I wouldn't say so

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -12,7 +12,7 @@ declare module "fastify" {
     destroySession(callback: (err?: Error) => void): void;
   }
 
-  interface Session extends Record<string, any> {
+  interface Session {
     sessionId: string;
     encryptedSessionId: string;
     /** Updates the `expires` property of the session. */
--- a/README.md
+++ b/README.md
@@ -126,8 +126,9 @@ This plugin supports typescript, and you can extend fastify module to add your c
 
 declare module "fastify" {
     interface Session {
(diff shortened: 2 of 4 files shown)


types/types.d.ts
```diff
@@ -12,7 +12,7 @@ declare module "fastify" {
     destroySession(callback: (err?: Error) => void): void;
   }
 
-  interface Session extends Record<string, any> {
+  interface Session {
```
zekth, line 15: So this forces user to declare their own session struct like: 
```ts
declare module 'fastify' {
  interface Session {
    id?: number;
    foo: string;
  }
}

```

SimenB: yep 🙂

zekth: Nice. Can you just add the `id?: number` property in the example in the readme?

SimenB: [`89c375b` (#14)](https://github.com/fastify/session/pull/14/commits/89c375bc9df3d758728cbb3b870c117ee0c566af)?

zekth: Yes

zekth (approved):

climba03003 (approved):
LGTM.

PR #16, opened 2021-08-13
feat: set up session in `onRequest` rather than `preValidation`

SimenB:
Port of https://github.com/SerayaEryn/fastify-session/pull/147. Original OP below.

---

Being able to access the session data in any `onRequest` hook is useful (in my case _needed_ since we still use `fastify-express` which only runs in that hook).


climba03003 (approved):
LGTM.

zekth (approved):
Lgtm. Thanks!

Issue #21, opened 2021-09-11
Move typescript stack to TSD


zekth:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

We need to move the typescript test stack to use `tsd` instead: https://github.com/SamVerschueren/tsd

### Motivation

Be inligned with all the other plugins in the core ecosystem

PR #22, opened 2021-09-13
feat: switch to tsd

zekth:
fix: #21


SimenB:
Why are most of the tests deleted instead of migrated?

zekth:
> Why are most of the tests deleted instead of migrated?

added missing assertions. But most of the test files were compilations tests aren't they? Do i miss something in here?

SimenB:
Yeah, seems to have been quite a bit of duplication

types/test/basic.ts
```diff
@@ -1,58 +0,0 @@
...
-import cookie from 'fastify-cookie';
-import session from '../../';
-
-const app = fastify();
-
-declare module 'fastify' {
-  interface Session {
-    id?: number;
-  }
-}
```
SimenB, line 11: this part is missing - adding own properties to the session object

types/test/basic.ts
```diff
@@ -1,58 +0,0 @@
...
-  async handler(_req, _rep) {
-    return { status: 'ok' };
-  },
-});
-
-app.route({
-  method: 'GET',
-  url: '/logout',
-  preHandler(req, _rep, next) {
-    req.destroySession(next);
```
SimenB, line 36: this function is not asserted on

types/test/session-type.ts
```diff
@@ -1,65 +0,0 @@
...
-import session from '../../';
-
-declare module 'fastify' {
-  interface Session {
-    user?: {
-      id: number;
-    };
-  }
-}
-
```
SimenB, line 12: same here I guess, but a deeper object. Both should get tests I guess?

types/types.d.ts
```diff
@@ -92,10 +92,7 @@ declare namespace FastifySessionPlugin {
 export class MemoryStore implements FastifySessionPlugin.SessionStore {
   constructor(map?: Map<string, any>);
   set(sessionId: string, session: any, callback: (err?: Error) => void): void;
```
SimenB, line 94: why is `session` typed as `any` instead of `Session`?

SimenB (changes requested):
We need to assert the interface merging of the session object works, beyond that LGTM 👍

types/types.d.ts
zekth: This is inherited from previous code. We can fix it in another PR.

mcollina (approved):
lgtm

types/types.test-d.ts
```diff
@@ -0,0 +1,67 @@
...
+
+  destroy(_sessionId: string, _callback: Function) {}
+}
+
+declare module "fastify" {
+  interface Session {
+    sessionId: string;
+    encryptedSessionId: string;
+    touch(): void;
+    regenerate(): void;
```
SimenB, line 23:
```suggestion
```

should merge

SimenB (approved):

Issue #24, opened 2021-09-23
Promise API


SimenB:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Currently `SessionStore` (and `request.destroySession`) are callback APIs. Ideally these should be Promise-based instead (or support both if you want to keep that).

Since adaptors (such as `connect-redis`) all expect callback APIs, that API probably has to be kept at least as backing implementation. So dual API (like Fastify itself) probably makes sense.

### Motivation

It's more ergonomic, especially in larger async functions.

### Example

Instead of

```js
await promisify(sessionStore.set.bind(sessionStore))(
  session.sessionId,
  session,
);

//

await promisify(sessionStore.destroy.bind(sessionStore))(session.sessionId);
```

we could do

```js
await sessionStore.set(session.sessionId, session);

//

await sessionStore.destroy(session.sessionId);
```

zekth:
Let's keep both if possible. Would you work on this?

SimenB:
I'm happy to work on this at some point in the future, but I cannot predict at what point I'll have time for it. So if somebody else is up for it that'd be great 🙂

PR #31, opened 2021-10-29
feat: support promise api

lokshunhung:
Reolves #24: This PR adds promise support for `FastifyRequest#destroySession` and the `get`, `set`, `destroy` methods on `FastifyInstance#sessionStore`.

The session store methods are wrapped with a new function to inspect whether a callback argument is passed, and will return a promise if the callback is missing. This approach does not mutate/monkey-patch methods on the store instance to avoid breaking anything. Tests for this are included [here](https://github.com/lokshunhung/session/blob/fb7e7ea003d7ffcef4d75aee60c47cfb66c0ac15/test/proxy-store.test.js#L42-L114).

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,10 +96,40 @@ Function used to generate new session IDs. Defaults to [`uid(24)`](https://githu
 
 Allows to access or modify the session data.
 
+#### request.sessionStore
+
+Access the store instance from `options.store`. It proxies the `get`, `set`, `destroy` method on the store instance, such that it returns a promise when the callback argument is not passed. Other methods and properties on the store instance can be accessed transparently.
+
+```ts
+// Using callbacks
+request.sessionStore.get(sessionId, (err, session) => {})
+request.sessionStore.set(sessionId, session, (err) => {})
+request.sessionStore.destroy(sessionId, (err) => {})
+// Using promises
+request.sessionStore.get(sessionId).then((session) => {}).catch((err) => {})
+request.sessionStore.set(sessionId, session).catch((err) => {})
+request.sessionStore.destroy(sessionId).catch((err) => {})
+// Using async/await
+const session = await request.sessionStore.get(sessionId)
+await request.sessionStore.set(sessionId, session)
+await request.sessionStore.destroy(sessionId)
+// Some store implementations has extra methods, they can be used normally
+request.sessionStore.touch(sessionId, session, (err) => {})
+```
+
 #### request.destroySession(callback)
 
 Allows to destroy the session in the store.  
 If callback is not passed, request.destroySession returns a promise that will reject if an error has occured.
+
+```ts
+// Using callbacks
+request.destroySession((err) => {})
+// Using promises
+request.destroySession.catch((err) => {})
+// Using async/await
+await request.destroySession()
+```
 
 #### Session#touch()
 


lokshunhung:
@mcollina Thanks for taking a look, unfortunately i have misunderstood the issue. Is there any existing promise-based store packages that the library would like to support? I could use it as reference and work on it. Cheers

mcollina:
@SimenB wdyt? Do you know anything?

SimenB:
What I originally requested was a way to use promises instead of callbacks in the API exposed by `@fastify/session`. Since this module is made for compat with Express' stores, I don't think any of those being promise based is anywhere close to happening.

So what I wanted is essentially what this PR has implemented - wrap any callback APIs with promises.

(as mentioned there, `request.destroySession` is also a callback, but all other methods today are from the stores which are plugged in)

lokshunhung:
@mcollina I can continue to work on it if you think this could be something to be added to the library. Currently the CI fails due to coverage which should not be too hard to fix.

mcollina:
Great news! Could you fix the remaining tests @lokshunhung ?

lokshunhung:
Sure, I can take a look next morning. It's 12am now so good night guys.

lokshunhung:
@mcollina I fixed the tests and updated the docs. It's ready now.

README.md
```diff
@@ -74,7 +74,7 @@ The options object is used to generate the `Set-Cookie` header of the session co
 * `domain` - The `Domain` attribute.
 
 ##### store
-A session store. Needs the following methods: 
+A session store. The store should be an event emitter with the following methods:
```
zekth, line 77: I don't get the change. It's not an event emitter.

lokshunhung: Actually [it's an event emitter](https://github.com/fastify/session/blob/4546ff96cd3de71a4f33e6b11522adc5f8a3ad6b/lib/store.js#L8), and there is a [test](https://github.com/fastify/session/blob/4546ff96cd3de71a4f33e6b11522adc5f8a3ad6b/test/store.test.js#L109-L116) for this.

While I do not fully understand why it needs to be an event emitter (since the event emitter API is not used anywhere), I believe that this originated from `express-session` to [track if the store is disconnected](https://github.com/expressjs/session/blob/0048bcac451ad867299d404aca94c79cc8bc751d/index.js#L188), and I guess this library aiming for compatibility with `express-session` stores ported it, but did not implemented store connection tracking.

Do you think I should remove it?

zekth: Meanwhile this is due to the `express-session` compatibility it's not required to be an event emitter. I feel it's just a bootstrap for the direct compatibility. Only the described methods are required.
I feel it might be misinterpreted by end users. Might be worth adding a comment in the code for this (in another PR)

lokshunhung: > might be misinterpreted by end users

You're right. I'll remove it.

lib/destroy-session.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+
+module.exports = function destroySession (done) {
```
SimenB, line 3: an alternative approach to this PR is to do what this function does (put the store's methods directly on `request` and encapsulate the session ID), and make `sessionStore` remain what the user passed in (i.e. callback only).

e.g. `request.sessionStore.set(request.session.sessionId, {})` could just be `request.setSession({})`, `request.sessionStore.get(request.session.sessionId)` could be `request.getSession()`.

SimenB (approved):
late suggestion for alternative approach, but current code works for mye use cases 🙂

lib/destroy-session.js
SimenB: or even stick more methods on the `session` object itself rather than directly on the `request` object - today it has only `regenerate` and `touch`, but not `save` etc.. When using Express' `session` I never interacted with `sessionStore` directly, all methods (`save`, `regenerate`, `reload`, `destroy`) were all on the session object.

lokshunhung: @SimenB Thank you for the input! My only gripe with putting methods on the session is that it "occupies" the names that you use on `request.session`. While it is technically possible to do `request.session.destroy = someValue` (since the methods live on the session class's prototype object), it should be avoided as it causes confusion.

Side note:  
I like the approach that [fastify-secure-session](https://github.com/fastify/fastify-secure-session) went with, that to set session values you must use `request.session.set('name', 'value')`. Which enables keeping track if the session is changed with a simple [flag](https://github.com/fastify/fastify-secure-session/blob/ee17469d15472614b47f9390c806cd9329b4a6f1/index.js#L204) and avoid [this](https://github.com/fastify/session/blob/4546ff96cd3de71a4f33e6b11522adc5f8a3ad6b/lib/fastifySession.js#L211). But this means a breaking change on the session interface.

SimenB: `touch` and `regenerate` is already on `session`, as my table shows the current API is _very_ inconsistent.

(I personally have nothing against going through a setter rather than using assignment, hugely breaking for this module, tho)

lokshunhung: That is indeed a huge breaking change, and I'm not sure if end users of this library would appreciate. I think it's in everyone's best interest to fix the inconsistency without breaking too many things, or at least having a clear way on migrating if we decide that the setter syntax is the way to go.

If this library uses the setter syntax that would also solve [this issue](https://github.com/fastify/fastify-passport/issues/319) from fastify-passport. I had been trying to make fastify-passport work with a storage-backed session middleware (with the goal of migrating projects from express to fastify) so I started building [this](https://github.com/lokshunhung/fastify-compat-session) but did not have the time to finish it. So I guess it's better to pool efforts collectively from the community than to roll things on my own.

lokshunhung: @mcollina What are your opinions on this?

climba03003: I think this topic can open a new issue for it. I support to unify the current API.
And it is not really related to this PR.

SimenB: Right, but by unifying this PR would probably be completely different

lokshunhung: I created a new issue for this: #34 

Can I mark this as resolved?

lokshunhung:
If the comment from @SimenB is resolved, I think this can be merged

SimenB (approved):
this works fine for me as is, but I do think teh API consistency should be solved first (to avoid potential churn)

zekth (approved):
Lgtm. Good work

mcollina:
>  I do think the API consistency should be solved first (to avoid potential churn)

Could you clarify what you mean by "first"?

SimenB:
> > I do think the API consistency should be solved first (to avoid potential churn)
> 
> Could you clarify what you mean by "first"?

Before landing this PR, as I think a unified API will remove the reason to promisify `sessionStore` in the first place

mcollina:
@SimenB so you prefer if https://github.com/fastify/session/issues/34 was solved first and then this to be reworked on top. Great.

SimenB:
Yup!

PR #50, opened 2021-12-21
Add Get/Set methods

sameer-coder:
This PR adds `get`, `set` methods for using the session. 
These methods are need for integration with `fastify-passport` (See https://github.com/fastify/fastify-passport/pull/421)

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -19,6 +19,10 @@ declare module "fastify" {
     touch(): void;
     /** Regenerates the session by generating a new `sessionId`. */
     regenerate(): void;
+    /** sets values in the session. */
+    set(key: string, value: any): void;
+    /** gets values from the session. */
+    get(key: string): any;
   }
 }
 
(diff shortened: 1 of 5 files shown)


types/types.d.ts
```diff
@@ -19,6 +19,10 @@ declare module "fastify" {
     touch(): void;
     /** Regenerates the session by generating a new `sessionId`. */
     regenerate(): void;
+    /** sets values in the session. */
+    set(key: string, value: any): void;
```
zekth, line 23: can't we use unknown in here?

types/types.d.ts
```diff
@@ -19,6 +19,10 @@ declare module "fastify" {
...
+    /** gets values from the session. */
+    get(key: string): any;
```
zekth, line 25: is something like `get<T>(key: string): T` possible?

zekth (changes requested):
Questions about types. Also could you please add types tests in: https://github.com/fastify/session/blob/f8452014d8a2576adc340e743cb93731cc230979/types/types.test-d.ts

zekth (approved):
LGTM :v:

mcollina (approved):
lgtm

climba03003 (approved):
LGTM.

Issue #62, opened 2022-03-06
Custom ID generator request parameter


Legolaszstudio:
### Prerequisites

- [X] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

idGenerator should pass the request parameter to the generator function.

### Motivation

Coming from express-session this seems a reasonable feature, as we assign a prefix to users logged in.

### Example

```ts
fastify.register(fastifySession, {
    store: new RedisStore({ client: redisClient }) as any,
    secret: 'abc123', 
    idGenerator: (req) => {
        if (req.authenticated) return `user_${req.userUUID}:${uuidv1()}`;
        else return uuidv1();
    },
    cookie: {
        httpOnly: true,
    },
});
```

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

Legolaszstudio:
I have a lot of things coming up the following weeks, but I will try my best 😀

Legolaszstudio:
Fixed by #64

PR #64, opened 2022-03-09
idGenerator request object (#62)

Legolaszstudio:
#### Checklist

#### What did I do?

Fix for #62 
Added the request object to the idGenerator function for custom scenarios like:
```js
idGenerator: (request) => {
     if (request.session.returningVisitor) return `returningVisitor-${uid(24)}`
     else return uid(24)
}
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ Updates the `expires` property of the session.
 
 #### Session#regenerate(request)
 
-Regenerates the session by generating a new `sessionId`. Don't forget to pass the fastify.request object if your id generator uses it.
+Regenerates the session by generating a new `sessionId`. Don't forget to pass the `request` object if the `idGenerator` function parameter uses it.
 ```js
 fastify.get('/regenerate', (request, reply) => {
   request.session.regenerate(request);
--- a/README.md
+++ b/README.md
@@ -88,9 +88,16 @@ Defaults to a simple in-memory store.</br>
 Save sessions to the store, even when they are new and not modified— defaults to `true`.
 Setting this to `false` can save storage space and comply with the EU cookie law.
 
-##### idGenerator (optional)
+##### idGenerator(request) (optional)
 
 Function used to generate new session IDs. Defaults to [`uid(24)`](https://github.com/crypto-utils/uid-safe).
+Custom implementation example:
+```js
+idGenerator: (request) => {
+     if (request.session.returningVisitor) return `returningVisitor-${uid(24)}`
+     else return uid(24)
+}
+```
 
 #### request.session
 
@@ -104,9 +111,15 @@ Allows to destroy the session in the store
 
 Updates the `expires` property of the session.
 
-#### Session#regenerate()
+#### Session#regenerate(request)
 
-Regenerates the session by generating a new `sessionId`.
+Regenerates the session by generating a new `sessionId`. Don't forget to pass the fastify.request object if your id generator uses it.
+```js
+fastify.get('/regenerate', (request, reply) => {
+  request.session.regenerate(request);
+  reply.send(request.session.sessionId);
+});
+```
 
 #### Session#get(key)
 
--- a/lib/session.js
+++ b/lib/session.js
@@ -10,7 +10,7 @@ const addDataToSession = Symbol('addDataToSession')
 const generateId = Symbol('generateId')
 
 module.exports = class Session {
-  constructor (idGenerator, cookieOpts, secret, prevSession = {}) {
+  constructor (request, idGenerator, cookieOpts, secret, prevSession = {}) {
     this[generateId] = idGenerator
     this.expires = null
     this.cookie = new Cookie(cookieOpts)
@@ -19,7 +19,7 @@ module.exports = class Session {
     this[addDataToSession](prevSession)
     this.touch()
     if (!this.sessionId) {
-      this.regenerate()
+      this.regenerate(request)
     }
   }
 
@@ -30,8 +30,8 @@ module.exports = class Session {
     }
   }
 
-  regenerate () {
-    this.sessionId = this[generateId]()
+  regenerate (request) {
+    this.sessionId = this[generateId](request)
     this.encryptedSessionId = this[sign]()
   }
 
@@ -55,8 +55,8 @@ module.exports = class Session {
     return cookieSignature.sign(this.sessionId, this[secretKey])
   }
 
-  static restore (idGenerator, cookieOpts, secret, prevSession) {
-    const restoredSession = new Session(idGenerator, cookieOpts, secret, prevSession)
+  static restore (request, idGenerator, cookieOpts, secret, prevSession) {
+    const restoredSession = new Session(request, idGenerator, cookieOpts, secret, prevSession)
     const restoredCookie = new Cookie(cookieOpts)
     restoredCookie.expires = new Date(prevSession.cookie.expires)
     restoredSession.cookie = restoredCookie
--- a/test/session.test.js
+++ b/test/session.test.js
@@ -353,18 +353,8 @@ test('should use custom sessionId generator if available (with request)', async 
     secret: 'cNaoPYAwF60HZJzkcNaoPYAwF60HZJzk',
     cookie: { secure: false, maxAge: 10000 },
     idGenerator: (request) => {
-      if (request.session && request.session.returningVisitor) {
-        return `returningVisitor-${
-          new Date().getTime()
-        }-${
-          Math.random().toString().slice(2)
-        }`
-      }
-      return `custom-${
-        new Date().getTime()
-      }-${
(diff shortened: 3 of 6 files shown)


types/types.d.ts
```diff
@@ -27,7 +27,7 @@ interface SessionData extends ExpressSessionData {
   touch(): void;
 
   /** Regenerates the session by generating a new `sessionId`. */
-  regenerate(): void;
```
Legolaszstudio, line 30: Should this be forced like this or should I change to a nullable parameter?

types/types.d.ts
```diff
@@ -27,7 +27,7 @@ interface SessionData extends ExpressSessionData {
...
+  regenerate(request: Fastify.FastifyRequest): void;
```
Legolaszstudio, line 30: Should I change to a nullable or keep it like this?

Legolaszstudio:
I wrote a very simple test case, and in my opinion, there isn't really anything else to test.
In case there is an edge case or something like that, it won't really cover that....

lib/session.js
```diff
@@ -55,8 +55,8 @@ module.exports = class Session {
     return cookieSignature.sign(this.sessionId, this[secretKey])
   }
 
-  static restore (idGenerator, cookieOpts, secret, prevSession) {
-    const restoredSession = new Session(idGenerator, cookieOpts, secret, prevSession)
+  static restore (request, idGenerator, cookieOpts, secret, prevSession) {
```
mcollina, line 58: Is this change tested?

types/types.d.ts
mcollina: cc @fastify/typescript

mcollina (changes requested):
Could you please add a test for the types? We use tsd.

darkgl0w (changes requested):
Some changes and types unit tests are missing.

test/session.test.js
```diff
@@ -345,3 +345,58 @@ test('should update the expires property of the session using Session#touch() ev
...
 })
+
+test('should use custom sessionId generator if available (with request)', async (t) => {
+  const fastify = Fastify()
+  fastify.register(fastifyCookie)
+  fastify.register(fastifySession, {
+    secret: 'cNaoPYAwF60HZJzkcNaoPYAwF60HZJzk',
+    cookie: { secure: false, maxAge: 10000 },
+    idGenerator: (request) => {
+      if (request?.session.returningVisitor) {
```
darkgl0w, line 356: Optional chaining is not supported by Node.js 10

types/types.d.ts
Legolaszstudio: Ok, let's make it optional, so we don't break anything.
But make sure that it is clearly documented.

test/session.test.js
Legolaszstudio: Fixed in
https://github.com/fastify/session/pull/64/commits/99614dad9a9cd78d6bc0daec75ba64d4723ebff9

types/types.d.ts
```diff
@@ -101,7 +101,7 @@ declare namespace FastifySessionPlugin {
     rolling?: boolean;
 
     /** Function used to generate new session IDs. Defaults to uid(24). */
-    idGenerator?: () => string;
+    idGenerator?: (request: Fastify.FastifyRequest) => string;
```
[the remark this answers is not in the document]
Legolaszstudio: It is optional indeed, my mistake

lib/session.js
Legolaszstudio: Restore is only referenced once; To be exact in fastifySession line 80 (https://github.com/fastify/session/pull/64/files#diff-c327147bd50b6dd851469add08b782e4ede4da6003ddbab99f46e50ff220d84bR80-R84)

I don't see how this could be tested? I only added a request parameter to it which will be called inside Session's constructor which is tested in theory. Nevertheless I will add a rolling false custom id generator test case

Legolaszstudio:
Fixed everything mentioned, except:
> Could you please add a test for the types? We use tsd.

How do I test the types and what should I test on them?

types/types.d.ts
```diff
@@ -26,8 +26,11 @@ interface SessionData extends ExpressSessionData {
   /** Updates the `expires` property of the session. */
   touch(): void;
 
-  /** Regenerates the session by generating a new `sessionId`. */
-  regenerate(): void;
+  /**
+   * Regenerates the session by generating a new `sessionId`.
+   * @param request Optional, but REQUIRED if idGenerator uses it
+   */
+  regenerate(request?: Fastify.FastifyRequest): void;
```
darkgl0w, line 33:
```suggestion
  /**
   * Regenerates the session by generating a new `sessionId`.
   * @param request optional, but __REQUIRED__ if the `idGenerator` function used to generate new session IDs uses it.
   */
  regenerate(request?: Fastify.FastifyRequest): void;
```

darkgl0w: One other way to type the `regenerate` method we can explore would be something like this:
```ts
// With this we can get the parameters of a function type in a tuple
type Parameters<T> = T extends (...args: infer P) => any ? P : never;

// and by typing regenerate like this it will accept the same arguments (and types) that `idGenerator` accepts
// NB: if we add another argument to `idGenerator` it will be added to `regenerate` too.
regenerate(...args: Parameters<FastifySessionPlugin.Options['idGenerator']>): void;
```

cc @fastify/typescript WDYT ?

darkgl0w (changes requested):
A few nitpicks and typescript definitions tests are missing.

Take a look at the `types/types.test-d.ts` to see how we test our typescript definitions. ^^
You need to add a case where you register the plugin with an `idGenerator` function that takes a `request` argument and uses it to generate a `sessionId`, and by extension a `regenerate` use case that takes a `request` argument too.

README.md
```diff
@@ -104,9 +111,15 @@ Allows to destroy the session in the store
 
 Updates the `expires` property of the session.
 
-#### Session#regenerate()
+#### Session#regenerate(request)
 
-Regenerates the session by generating a new `sessionId`.
+Regenerates the session by generating a new `sessionId`. Don't forget to pass the fastify.request object if your id generator uses it.
```
Eomm, line 116:
```suggestion
Regenerates the session by generating a new `sessionId`. Don't forget to pass the `request` object if the `idGenerator` function parameter uses it.
```

test/session.test.js
```diff
@@ -345,3 +345,114 @@ test('should update the expires property of the session using Session#touch() ev
...
+  fastify.register(fastifySession, {
+    secret: 'cNaoPYAwF60HZJzkcNaoPYAwF60HZJzk',
+    cookie: { secure: false, maxAge: 10000 },
+    idGenerator: (request) => {
+      if (request.session && request.session.returningVisitor) {
+        return `returningVisitor-${
+          new Date().getTime()
+        }-${
+          Math.random().toString().slice(2)
+        }`
```
Eomm, line 361: Could you split this statement into multiple ones?

test/session.test.js
```diff
@@ -345,3 +345,114 @@ test('should update the expires property of the session using Session#touch() ev
...
+      }
+      return `custom-${
+        new Date().getTime()
+      }-${
+        Math.random().toString().slice(2)
+      }`
```
Eomm, line 367: Could you split this statement into multiple ones? (it is hard to read and there is some duplicated code

types/types.test-d.ts
```diff
@@ -50,6 +50,10 @@ app.register(plugin, {
   secret: 'ABCDEFGHIJKLNMOPQRSTUVWXYZ012345',
   idGenerator: () => Date.now() + ''
 });
+app.register(plugin, {
+  secret: 'ABCDEFGHIJKLNMOPQRSTUVWXYZ012345',
+  idGenerator: (request) => request?.ip + '-' + Date.now()
```
darkgl0w, line 55: Node.js 10 does not support optional chaining ^^

darkgl0w (approved):
lgtm

Legolaszstudio:
I think everything requested is done, only waiting for
> One other way to type the `regenerate` method we can explore would be something like this:
> 
> ```ts
> // With this we can get the parameters of a function type in a tuple
> type Parameters<T> = T extends (...args: infer P) => any ? P : never;
> 
> // and by typing regenerate like this it will accept the same arguments (and types) that `idGenerator` accepts
> // NB: if we add another argument to `idGenerator` it will be added to `regenerate` too.
> regenerate(...args: Parameters<FastifySessionPlugin.Options['idGenerator']>): void;
> ```
> 
> cc @fastify/typescript WDYT ?
> 
> Edit: it seems that Typescript already has a `Parameters` utility defined like this:
> 
> ```ts
> type Parameters<T extends (...args: any) => any> = T extends (...args: infer P) => any ? P : never
> ```

mcollina (approved):
lgtm

Issue #52, opened 2022-01-07
Exposing encryptedSessionId leaks secret to database


cameronmbell:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

3.25.3

### Plugin version

6.3.0

### Node.js version

14.15.1

### Operating system

Windows

### Operating system version (i.e. 20.04, 11.3, 10)

10

### Description

Compared to express session, fastify session exposes the encrypted session id to the database. This could be used to determine the session secret given a database leak and thus leaking temporary password equivalent. It also wastes space by including the sessionId twice and expiry date three times.

express-session
```json
"_id":"abcdefghijklmnopqrstuvwxyz",
"expires":"2022-03-08T09:43:51.997+00:00",
"session":{
    "cookie":{
        "originalMaxAge":"5184000000",
        "expires":"2022-03-08T09:43:35.786+00:00",
        "secure":false,
        "httpOnly":true,
        "domain":null,
        "path":"/",
        "sameSite":null
    }
}
```
fastify/session
```json
"_id":"abcdefghijklmnopqrstuvwxyz",
"expires":"2022-03-08T09:43:51.997+00:00",
"session":{
    "expires":"2022-03-08T09:43:51.997+00:00",
    "cookie":{
        "originalMaxAge":"5184000000",
        "expires":"2022-03-08T09:43:35.786+00:00",
        "secure":false,
        "httpOnly":true,
        "domain":null,
        "path":"/",
        "sameSite":"strict"
    },
    "sessionId":"abcdefghijklmnopqrstuvwxyz",
    "encryptedSessionId":"qwertyuiopasdfghjklzxcvbnm"
}
```

### Steps to Reproduce

```js
import fp from 'fastify-plugin';
import cookie from 'fastify-cookie';
import session from '@fastify/session';

// https://github.com/jdesboeufs/connect-mongo/issues/441
import _ from 'express-session';
import MongoStore from 'connect-mongo';

export default fp(async (fastify, options) => {
    if (!fastify.dbClient) {
        throw new Error('Must register db plugin');
    }

    fastify.register(cookie);
    fastify.register(session, {
        secret: process.env.SECRET,
        saveUninitialized: false,
        cookieName: '__Host-sessionId',
        cookie: {
            sameSite: 'strict',
            httpOnly: true,
            secure: false
        },
        store: MongoStore.create({
            stringify: false,
            client: /* ... */
        })
    });
});
```

[...]

PR #70, opened 2022-03-23
Do not leak encryptedSessionId to session

m0dch3n:
#### Checklist

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -112,6 +112,7 @@ function onRequest (options) {
 }
 
 function onSend (options) {
+  const secret = options.secret[0]
   return function saveSession (request, reply, payload, done) {
     const session = request.session
     if (!session || !session.sessionId || !shouldSaveSession(request, options.cookie, options.saveUninitialized)) {
@@ -123,9 +124,10 @@ function onSend (options) {
         done(err)
         return
       }
+      const encryptedSessionId = cookieSignature.sign(session.sessionId, secret)
       reply.setCookie(
         options.cookieName,
-        session.encryptedSessionId,
+        encryptedSessionId,
         session.cookie.options(isConnectionSecure(request))
       )
       done()
@@ -218,7 +220,7 @@ function shouldSaveSession (request, cookieOpts, saveUninitialized) {
 }
 
 function isSessionModified (session) {
-  return (Object.keys(session).length !== 4)
+  return (Object.keys(session).length !== 3)
 }
 
 function option (options, key, def) {
--- a/lib/session.js
+++ b/lib/session.js
@@ -1,11 +1,9 @@
 'use strict'
 
 const Cookie = require('./cookie')
-const cookieSignature = require('cookie-signature')
 
 const maxAge = Symbol('maxAge')
 const secretKey = Symbol('secretKey')
-const sign = Symbol('sign')
 const addDataToSession = Symbol('addDataToSession')
 const generateId = Symbol('generateId')
 
@@ -32,7 +30,6 @@ module.exports = class Session {
 
   regenerate (request) {
     this.sessionId = this[generateId](request)
-    this.encryptedSessionId = this[sign]()
   }
 
   [addDataToSession] (prevSession) {
@@ -49,10 +46,6 @@ module.exports = class Session {
 
   set (key, value) {
     this[key] = value
-  }
-
-  [sign] () {
-    return cookieSignature.sign(this.sessionId, this[secretKey])
   }
 
   static restore (request, idGenerator, cookieOpts, secret, prevSession) {
--- a/test/base.test.js
+++ b/test/base.test.js
@@ -119,7 +119,7 @@ test('should set session cookie using the default cookie name', async (t) => {
   })
 
   t.is(statusCode, 200)
-  t.regex(cookie, /sessionId=undefined; Path=\/; HttpOnly; Secure/)
+  t.regex(cookie, /sessionId=.*\..*; Path=\/; HttpOnly; Secure/)
 })
 
 test('should create new session on expired session', async (t) => {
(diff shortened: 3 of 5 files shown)


m0dch3n:
Fixes: #52

mcollina (approved):
lgtm

zekth (approved):
thanks! Is it considered as breaking change?

darkgl0w (approved):

darkgl0w:
> thanks! Is it considered as breaking change?

Yes IMO it should be considered a breaking change as it changes the public API of the package and it changes the "user land" encrypted session id management. ^^

lib/fastifySession.js
```diff
@@ -218,7 +220,7 @@ function shouldSaveSession (request, cookieOpts, saveUninitialized) {
 }
 
 function isSessionModified (session) {
-  return (Object.keys(session).length !== 4)
+  return (Object.keys(session).length !== 3)
```
jsumners, line 223: Session is not a plain JavaScript object. It is an instance of https://github.com/fastify/session/blob/73a4675b795e8ad3d020d5cd43de78374c09f8aa/lib/session.js#L12

That object should provide a `.keys()` method to get this information correctly.

lib/session.js
```diff
@@ -32,7 +30,6 @@ module.exports = class Session {
 
   regenerate (request) {
     this.sessionId = this[generateId](request)
-    this.encryptedSessionId = this[sign]()
```
SimenB, line 35: instead of removing it from `Session`, what about just ensuring it's not saved to the store? I.e. defining a custom `toJSON` or something

mcollina:
@SimenB could you review?
@m0dch3n could you rebase?

SimenB:
I think what would be most useful would be for the `sign` method on `Session` to become public. That way it's still easy to access if needed, but it's not stored in the session store

Eomm:
Sorry for being late, but there are conflicts here

jsumners:
@m0dch3n are you able to update this PR to resolve the conflicts?

SimenB:
Was finally able to pick up our migration from `express-session` to this module again, and _sorta_ encountered this. Since the _existing_ session (from `express-session`) in storage does _not_ have `sessionId` (and `encryptedSessionId`, but that doesn't make a difference in this context) a new ID is generated (https://github.com/fastify/session/blob/76459c2734a5bbe25cdc1cc86f0e454b9302219c/lib/session.js#L32-L35) and we "lose" the existing session since the ID is different from the one in storage.

In our own code I've patched `decryptSession` to pass through the `sessionId` it looked up in the store and pass it through to `new Session` and `Session.restore`, but ideally we'd figure out a way to remove both `sessionId` and `encryptedSessionId` from the session data itself. It's up to each store to translate id to storage key - `@fastify/session` shouldn't store the key itself in its data.

I think making both the id and the encrypted id getters should work?

FWIW, this is the patch I've got locally

<details>

```diff
diff --git a/lib/fastifySession.js b/lib/fastifySession.js
index b158eb83372c7c5125241e77a03ae575d17c5b1b..eda323c1e6fed788ff181d9110a0e64252d8bb24 100644
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -88,7 +88,8 @@ function decryptSession (sessionId, options, request, done) {
           idGenerator,
           cookieOpts,
           secret,
-          session
+          session,
+          decryptedSessionId
         )
       } else {
         request.session = Session.restore(
@@ -96,7 +97,8 @@ function decryptSession (sessionId, options, request, done) {
           idGenerator,
           cookieOpts,
           secret,
-          session
+          session,
+          decryptedSessionId
         )
       }
       done()
diff --git a/lib/session.js b/lib/session.js
index 59eca9ccc5df739d9a6496148ee88fd0af23c785..8067ab16779ea63b564f6ebee145bbf5b99ada3d 100644
--- a/lib/session.js
+++ b/lib/session.js
@@ -19,7 +19,7 @@ const originalHash = Symbol('originalHash')
 const hash = Symbol('hash')
 
 module.exports = class Session {
-  constructor (request, idGenerator, cookieOpts, secret, prevSession = {}) {
+  constructor (request, idGenerator, cookieOpts, secret, prevSession = {}, sessionId) {
     this[generateId] = idGenerator
     this.expires = null
     this.cookie = new Cookie(cookieOpts)
@@ -30,7 +30,11 @@ module.exports = class Session {
     this[requestKey] = request
     this.touch()
     if (!this.sessionId) {
-      this.sessionId = this[generateId](this[requestKey])
+      if (sessionId) {
+        this.sessionId = sessionId
+      } else {
+        this.sessionId = this[generateId](this[requestKey])
+      }
       this.encryptedSessionId = this[sign]()
     }
     this[originalHash] = this[hash]()
@@ -172,8 +176,8 @@ module.exports = class Session {
     return this[originalHash] !== this[hash]()
   }
 
-  static restore (request, idGenerator, cookieOpts, secret, prevSession) {
-    const restoredSession = new Session(request, idGenerator, cookieOpts, secret, prevSession)
+  static restore (request, idGenerator, cookieOpts, secret, prevSession, sessionId) {
+    const restoredSession = new Session(request, idGenerator, cookieOpts, secret, prevSession, sessionId)
     const restoredCookie = new Cookie(cookieOpts)
     restoredCookie.expires = new Date(prevSession.cookie.expires)
     restoredSession.cookie = restoredCookie
```
</details>

This is just to be able to migrate from `express-session` to `@fastify/session` - I do think ideally `@fastify/session` shouldn't keep `sessionId` directly on the session object passed to stores. It's quite useful that it exists on `request.session` though, and not `request.sessionID` like `express-session` does. Which is why I suggested a getter, but it might be a good idea to keep it separate?

mcollina:
Would you like to issue a fresh PR with your patch? I think a getter is a great idea too.

SimenB:
@mcollina opened #101 🙂

Issue #73, opened 2022-03-25
Breaking changes in SessionStore interface callback types


xcrzx:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the regression has not already been reported

### Last working version

6.4.0

### Stopped working in version

6.5.0

### Node.js version

17.4.0

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.3

### 💥 Regression Report

[This PR](https://github.com/fastify/session/pull/61/files#diff-98f0d704b0e4df307f4d90213bc415ce208d333a4b000524a8426ca733dd35a2) has introduced breaking changes to the public `SessionStore` interface by making all arguments non-optional in callbacks.

**Before:**
```ts
interface SessionStore {
    set(sessionId: string, session: Fastify.Session, callback: (err?: Error) => void): void;
    get(
      sessionId: string,
      callback: (err?: Error, session?: Fastify.Session) => void
    ): void;
    destroy(sessionId: string, callback: (err?: Error) => void): void;
  }
```

**After:**
```ts
  interface SessionStore {
    set(
      sessionId: string,
      session: Fastify.Session,
      callback: (err: Error) => void
    ): void;
    get(
      sessionId: string,
      callback: (err: Error, session: Fastify.Session) => void
    ): void;
    destroy(sessionId: string, callback: (err: Error) => void): void;
  }
```

This change is breaking and incorrect from the logic standpoint as calling the provided `callback` is required in session store implementation. And in case of success, there couldn't be any error to pass to the callback.

### Steps to Reproduce

See Regression Report

PR #61, opened 2022-03-01
fix (types): fix callbacks types and ensure compatibility with `express-session` stores

darkgl0w:
Hello.

This PR aims to :
- fix `CookieOptions` types
- avoid callbacks types bad practices
- make types compatible with `express-session` stores (see: https://github.com/fastify/help/issues/604)

#### Checklist

- [x] run `npm run test` ~~and `npm run benchmark`~~
- [x] tests ~~and/or benchmarks~~ are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #74, opened 2022-03-25
fix(types): error is optional, and session data always exists

SimenB:
#### Checklist

https://github.com/fastify/session/pull/61#issuecomment-1077885306

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -15,7 +15,7 @@ declare module 'fastify' {
     destroySession(callback: (err: Error) => void): void;
   }
 
-  interface Session extends Partial<SessionData> {}
+  interface Session extends SessionData {}
 }
 
 interface SessionData extends ExpressSessionData {
@@ -48,13 +48,13 @@ declare namespace FastifySessionPlugin {
     set(
       sessionId: string,
       session: Fastify.Session,
-      callback: (err: Error) => void
+      callback: (err?: Error) => void
     ): void;
     get(
       sessionId: string,
-      callback: (err: Error, session: Fastify.Session) => void
+      callback: (err: Error | null, session: Fastify.Session) => void
     ): void;
-    destroy(sessionId: string, callback: (err: Error) => void): void;
+    destroy(sessionId: string, callback: (err?: Error) => void): void;
   }
 
   interface Options {
@@ -136,13 +136,13 @@ export class MemoryStore implements FastifySessionPlugin.SessionStore {
   set(
     sessionId: string,
     session: Fastify.Session,
-    callback: (err: Error) => void
+    callback: (err?: Error) => void
   ): void;
   get(
     sessionId: string,
-    callback: (err: Error, session: Fastify.Session) => void
+    callback: (err: Error | null, session: Fastify.Session) => void
   ): void;
-  destroy(sessionId: string, callback: (err: Error) => void): void;
+  destroy(sessionId: string, callback: (err?: Error) => void): void;
 }
 
 export const Store: MemoryStore;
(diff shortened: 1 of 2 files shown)


types/types.d.ts
```diff
@@ -48,13 +48,13 @@ declare namespace FastifySessionPlugin {
     set(
       sessionId: string,
       session: Fastify.Session,
-      callback: (err: Error) => void
+      callback: (err?: Error) => void
     ): void;
     get(
       sessionId: string,
-      callback: (err: Error, session: Fastify.Session) => void
+      callback: (err: Error | null, session: Fastify.Session) => void
```
SimenB, line 55: I tried to use overloading so you either get an `err` _or_ a `session`, but I cannot make typescript understand that for whatever reason.

This is the way the core node types are typed, so probably fine, e.g.: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0beca137d8552f645064b8a622a6e153864c66ee/types/node/fs.d.ts#L898

darkgl0w: IMO that's the correct way to do it if we refer to the [typescript _Do's and Don'ts_ handbook section](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#overloads-and-callbacks).

types/types.d.ts
```diff
@@ -48,13 +48,13 @@ declare namespace FastifySessionPlugin {
     set(
       sessionId: string,
       session: Fastify.Session,
-      callback: (err: Error) => void
+      callback: (err?: Error) => void
```
darkgl0w, line 51: According to this [typescript Do's and Don'ts handbook section](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#optional-parameters-in-callbacks) callback parameters should not be optional as "it’s always legal to provide a callback that accepts fewer arguments".

But when looking at `express-session` typescript definitions, it's bloated with optional callbacks arguments and that makes it difficult to ensure compatibility between these packages when using typescript :exploding_head:

TBH I chose to follow the good practices of typescript here knowing that it could be controversial after looking at DefinitelyTyped types definitions for `express-session` and its compatible stores. :thinking:

darkgl0w (approved):
lgtm.
fix https://github.com/fastify/session/issues/73.

types/types.d.ts
SimenB: I made it optional to match the implementation

types/types.d.ts
xcrzx: The type of `callback` here should be `(err?: Error, session?: Fastify.Session) => void`. When retrieving a session, let's say from a database, you can get either `session` or `error`, but never both.

But the type `(err: Error | null, session: Fastify.Session) => void` makes session required. So the callback accept either session, or session + error.

types/types.d.ts
climba03003: `?: Error` do not really means optional parameter in this case, but it indicate that `error` may be `undefined`.
Generally speak, it should use `Error | undefined`, but `?: Error` is fewer character and look better.

types/types.d.ts
SimenB: again, that's not how node core types work

climba03003: Since callback always check with this pattern.
```js
function callback(err, session) {
  if(err) {
    // do something
    // ...
    return
  }
  
  // do somthing
  session 
}
```
If the session is `?: Fastify.Session` in the above case, the user will be inconvenient and need to double check the existence of `session`.

I would accept the current syntax.

climba03003 (approved):

PR #77, opened 2022-03-25
feat: add `destroy`, `save` and `reload` to session instance

### LINKED ISSUE #34: Unify inconsistent api for manipulating session
### Prerequisites

- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Originally proposed by @SimenB, as quoted

> [Add] methods on the `session` object itself rather than directly on the `request` object - today it has only `regenerate` and `touch`, but not `save` etc.. When using Express' `session` I never interacted with `sessionStore` directly, all methods (`save`, `regenerate`, `reload`, `destroy`) were all on the session object.
> 
> express	fastify
> `request.session.reload(cb)`	`request.sessionStore.get(request.session.sessionId, cb)`
> `request.session.save({}, cb)`	`request.sessionStore.set(request.session.sessionId, {}, cb)`
> `request.session.regenerate(cb)`	`request.session.regenerate()` (no callback in fastify, unlike express)
> `request.session.destroy(cb)`	`request.destroySession(cb)`
> `request.session.touch()`	`request.session.touch()`
> the methods for `express` also updates `request.session` - I guess calling the store directly doesn't do the same (that's probably a bug in my current migration at work).
> 
> If all methods are moved to `request.session` then I guess you can also choose not to expose `sessionStore` at all?

### Motivation

The api to interact with session is very inconsistent.

Context [here](https://github.com/fastify/session/pull/31#discussion_r740881005)

### Example

Instead of

```ts
request.sessionStore.get(request.session.sessionId, cb)
request.destroySession(cb)
```

We could do

```ts
await request.session.reload()
await request.session.destroy()
```

SimenB:
#### Checklist

I have not added tests yet, other than that I think this is ready to go. Opening it up early in case there's feedback before I spend time writing tests 🙂 

Fixes #34

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -28,7 +28,6 @@ function session (fastify, options, next) {
   })
   fastify.decorateRequest('sessionStore', { getter: () => options.store })
   fastify.decorateRequest('session', null)
-  fastify.decorateRequest('destroySession', destroySession)
   fastify.addHook('onRequest', onRequest(options))
   fastify.addHook('onSend', onSend(options))
   next()
@@ -65,7 +64,22 @@ function decryptSession (sessionId, options, request, done) {
         return
       }
       if (session && session.expires && session.expires <= Date.now()) {
-        options.store.destroy(sessionId, getDestroyCallback(secret, request, done, cookieOpts, idGenerator))
+        const restoredSession = Session.restore(
+          request,
+          idGenerator,
+          cookieOpts,
+          secret,
+          session
+        )
+
+        restoredSession.destroy(err => {
+          if (err) {
+            done(err)
+            return
+          }
+
+          restoredSession.regenerate(done)
+        })
         return
       }
       if (options.rolling) {
@@ -118,7 +132,7 @@ function onSend (options) {
       done()
       return
     }
-    options.store.set(session.sessionId, session, (err) => {
+    session.save((err) => {
       if (err) {
         done(err)
         return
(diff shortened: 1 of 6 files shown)


mcollina (commented):
looks good to me. However we should really look how we promisify all those methods, but that's a different PR.

SimenB:
Yeah, plan is to follow up with a PR promisifying these after landing. How do you want that done? Check for callback, and return promise if not passed?

lib/fastifySession.js
```diff
@@ -65,7 +64,22 @@ function decryptSession (sessionId, options, request, done) {
         return
       }
       if (session && session.expires && session.expires <= Date.now()) {
-        options.store.destroy(sessionId, getDestroyCallback(secret, request, done, cookieOpts, idGenerator))
+        const restoredSession = Session.restore(
```
SimenB, line 67: doing a `restore` here to have access to the methods

SimenB:
@mcollina tests pushed! 🙂

SimenB:
Thanks! I'll get started on a PR promisifying. Hopefully up today or tomorrow

zekth (approved):
Good work.
Could we bump the major version as it's a breaking change with : https://github.com/fastify/session/pull/70 and your promisifying PR too?

SimenB:
Sounds like a good plan to me 👍

types/types.test-d.ts
```diff
@@ -79,5 +78,8 @@ app.route({
     expectType<void>(request.session.set('foo', 'bar'));
     expectType<string>(request.session.get('foo'));
     expectType<void>(request.session.touch());
+    expectType<void>(request.session.reload(() => {}));
+    expectType<void>(request.session.destroy(() => {}));
+    expectType<void>(request.session.regenerate(() => {}));
```
SimenB, line 83:
```suggestion
    expectType<void>(request.session.regenerate(() => {}));
    expectType<void>(request.session.save(() => {}));
```

SimenB:
Promise PR: #78

PR #79, opened 2022-03-30
feat: remove stale cookies

SimenB:
#### Checklist

I noticed while using a memory storage that the old session cookie was kept around. Doesn't really hurt, but it confused me when I was debugging why my session wasn't working

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -128,7 +128,16 @@ function onRequest (options) {
 function onSend (options) {
   return function saveSession (request, reply, payload, done) {
     const session = request.session
-    if (!session || !session.sessionId || !shouldSaveSession(request, options.cookie, options.saveUninitialized)) {
+    if (!session || !session.sessionId) {
+      done()
+      return
+    }
+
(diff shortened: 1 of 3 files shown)


lib/fastifySession.js
```diff
@@ -128,7 +128,16 @@ function onRequest (options) {
...
-    if (!session || !session.sessionId || !shouldSaveSession(request, options.cookie, options.saveUninitialized)) {
+    if (!session || !session.sessionId) {
+      done()
+      return
+    }
+
+    if (!shouldSaveSession(request, options.cookie, options.saveUninitialized)) {
+      // if a session cookie is set, but has a different ID, clear it
+      if (request.cookies[options.cookieName] && request.cookies[options.cookieName] !== session.encryptedSessionId) {
+        reply.clearCookie(options.cookieName)
```
SimenB, line 139: I think a better approach would be to do the `clearCookie` call in https://github.com/fastify/session/blob/fbb54079a9b8609bcb141527711109b616f8ef48/lib/fastifySession.js#L62-L65 instead, but we don't have access to `reply` in there, and `decryptSession` is public API

mcollina (approved):
lgtm

PR #80, opened 2022-03-30
feat: only save session to storage when it changes

SimenB:
#### Checklist

The current `isSessionModified` function will return `true` if any property has ever been added to the session object, no matter if it has changed from what's in the store or not. This PR implements a hashing of the session content, and only saves to the store if it's changed. This is the same as `express-session` does.


jsumners:
Would it be more performant to extend this approach to inspect the value for a change before marking the session as changed? https://github.com/jsumners/fastify-server-session/blob/f02ac8dd30dd76705b01293a749f22e73e4649b4/lib/session.js#L7-L28

test/session.test.js
```diff
@@ -3,6 +3,7 @@
 const test = require('ava')
 const Fastify = require('fastify')
 const fastifyCookie = require('fastify-cookie')
+const sinon = require('sinon')
```
SimenB, line 6: I dunno what libraries you prefer for stubs, sinon seems fine?

test/session.test.js
```diff
@@ -676,3 +677,54 @@ test('save supports rejecting promises', async t => {
...
+    }
+  })
+
+  fastify.get('/', (request, reply) => {
+    request.session.userId = 42
+
+    reply.send(200)
+  })
+
+  const { statusCode: statusCode1, headers: headers1 } = await fastify.inject('/')
```
SimenB, line 713: I assume there are legacy reasons (i.e., the tests came with the fork) for not using `inject` in the other tests and this is fine to use?

lib/session.js
```diff
@@ -143,12 +148,34 @@ module.exports = class Session {
     return cookieSignature.sign(this.sessionId, this[secretKey])
   }
 
+  [hash] () {
+    const sess = this
+    const str = JSON.stringify(sess, function (key, val) {
```
mcollina, line 153: This is not a good hashing algorithm. How about using a stable algorithm? So two JS objects with properties in different orders will generate the same string.

lib/session.js
```diff
@@ -143,12 +148,34 @@ module.exports = class Session {
...
+      }
+
+      return val
+    })
+
+    return crypto
+      .createHash('sha1')
+      .update(str, 'utf8')
+      .digest('hex')
+  }
```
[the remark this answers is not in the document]
mcollina: I would use https://github.com/fastify/fastify-etag/blob/master/fnv1a.js instead. It's significantly faster.

mcollina (commented):
I'm not convinced This will actually improve the actual speed. Generating a JSON and hashing it is quite expensive.

lib/session.js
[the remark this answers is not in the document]
SimenB: happy to! would you be able to extract that into a separate module, or should I copy it over?

lib/session.js
SimenB: something like https://www.npmjs.com/package/fast-json-stable-stringify?

SimenB:
@mcollina I'd love to add a benchmark. I have one locally, but I'm not sure how realistic it is. Its results show that using the default store is faster than the approach in this, but _nobody_ should ever use the default store (memory store) in production, so that doesn't really prove anything. Using the [file store](https://www.npmjs.com/package/session-file-store) shows about a 4x improvement in performance using this branch and a redis store (probably most realistic of the three, and what we use at work) about 50%.

Of course calling the server 25 times in a tight loop is not realistic, but maybe this illustrates _something_? 😅

Difference comes from the fact we in this PR swap out `sessionStore.save` _and_ `reply.setCookie` with a single `hash` call (unless the session actually changes, which it does every 5 requests made in my benchmark).

Bench code:

```
╔═══════════════════════╤═════════╤═══════════════╤═══════════╗
║ Slower tests          │ Samples │        Result │ Tolerance ║
╟───────────────────────┼─────────┼───────────────┼───────────╢
║ old code (file)       │      24 │   1.52 op/sec │  ± 0.91 % ║
║ current code (file)   │      25 │   5.95 op/sec │  ± 3.17 % ║
║ old code (redis)      │      25 │  24.68 op/sec │  ± 1.45 % ║
║ current code (redis)  │      25 │  33.90 op/sec │  ± 2.11 % ║
║ current code (memory) │      25 │ 245.25 op/sec │  ± 3.73 % ║
╟───────────────────────┼─────────┼───────────────┼───────────╢
║ Fastest test          │ Samples │        Result │ Tolerance ║
╟───────────────────────┼─────────┼───────────────┼───────────╢
║ old code (memory)     │      25 │ 268.14 op/sec │  ± 3.90 % ║
╚═══════════════════════╧═════════╧═══════════════╧═══════════╝
```

```diff
diff --git c/bench.js w/bench.js
new file mode 100644
index 0000000..1e3ed21
--- /dev/null
+++ w/bench.js
@@ -0,0 +1,87 @@
+'use strict'
+
+const cronometro = require('cronometro')
+const Fastify = require('fastify')
+const fastifyCookie = require('fastify-cookie')
+const fileStoreFactory = require('session-file-store')
+const redisStoreFactory = require('connect-redis')
+const { createClient: createRedisClient } = require('redis')
+const fastifySession = require('.')
+const oldFastifySession = require('old-session')
+const fastifyPlugin = require('fastify-plugin')
+
+const sessionPlugin = async (fastify, { plugin, storeType }) => {
+  const FileStore = fileStoreFactory(plugin)
+  const RedisStore = redisStoreFactory(plugin)
+  const redisClient = createRedisClient({ legacyMode: true })
+  // only run this if you have redis running on localhost:6379
+  await redisClient.connect()
+
+  fastify.register(fastifyCookie)
+
+  fastify.register(plugin, {
+    secret: 'cNaoPYAwF60HZJzkcNaoPYAwF60HZJzk',
+    saveUninitialized: false,
+    cookie: { secure: false },
+    store: storeType === 'redis'
+      ? new RedisStore({ client: redisClient })
+      : storeType === 'file'
+        ? new FileStore({})
+        : undefined
+  })
+}
+
+const getServer = (plugin, storeType) => {
+  let requestCounter = 0
+
+  const fastify = Fastify()
+
+  fastify.register(fastifyPlugin(sessionPlugin), { plugin, storeType })
+
+  fastify.get('/', (request, reply) => {
+    // modify session every 10 requests
+    if (requestCounter % 10 === 0) {
+      request.session.userId = requestCounter
+    } else {
+      request.session.userId = 0
+    }
+
+    requestCounter++
+
+    reply.send(200)
+  })
+
+  return fastify
+}
+
+const testFunction = (sessionPlugin, storeType) => {
+  const server = getServer(sessionPlugin, storeType)
+
+  return async () => {
+    const { headers } = await server.inject('/')
+    const setCookieHeader = headers['set-cookie']
+
+    if (!setCookieHeader) {
+      throw new Error('Missing set-cookie header')
+    }
+
+    const { sessionId } = server.parseCookie(setCookieHeader)
+
+    // make 25 "requests" with the new session
+    for (let i = 0; i < 25; i++) {
+      await server.inject({ path: '/', headers: { cookie: `sessionId=${sessionId}` } })
+    }
+  }
+}
+
+cronometro(
+  {
+    'current code (memory)': testFunction(fastifySession),
+    'old code (memory)': testFunction(oldFastifySession),
+    'current code (file)': testFunction(fastifySession, 'file'),
+    'old code (file)': testFunction(oldFastifySession, 'file'),
+    'current code (redis)': testFunction(fastifySession, 'redis'),
+    'old code (redis)': testFunction(oldFastifySession, 'redis')
+  },
+  { iterations: 25 }
+)
diff --git c/package.json w/package.json
index 2a05309..9f90ca7 100644
--- c/package.json
+++ w/package.json
@@ -32,9 +32,14 @@
   "devDependencies": {
     "@types/node": "^17.0.0",
     "ava": "^4.1.0",
+    "connect-redis": "^6.1.3",
+    "cronometro": "^0.8.0",
     "fastify": "^3.0.0",
     "fastify-cookie": "^5.3.1",
     "nyc": "^15.0.0",
+    "old-session": "npm:@fastify/session@^8.0.0",
+    "redis": "^4.0.6",
+    "session-file-store": "^1.5.0",
     "sinon": "^13.0.1",
     "standard": "^16.0.1",
     "tsd": "^0.19.0",
```

lib/session.js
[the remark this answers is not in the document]
mcollina: You could also use the Sindre version, however that's esm-only.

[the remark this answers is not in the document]
mcollina: Extracting is ok, but you can copy over to verify.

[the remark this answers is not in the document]
SimenB: could we install an older version of https://github.com/sindresorhus/fnv1a that's not ESM only, or do you want to stay on latest? If we make this module ESM only we can no longer upgrade at work 😅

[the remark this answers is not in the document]
SimenB: copying over gave this bench result (same as in below comment)

```
╔═══════════════════════╤═════════╤═══════════════╤═══════════╗
║ Slower tests          │ Samples │        Result │ Tolerance ║
╟───────────────────────┼─────────┼───────────────┼───────────╢
║ old code (file)       │       6 │   1.52 op/sec │  ± 0.76 % ║
║ current code (file)   │      25 │   6.24 op/sec │  ± 2.73 % ║
║ old code (redis)      │      25 │  24.97 op/sec │  ± 1.03 % ║
║ current code (redis)  │      25 │  36.58 op/sec │  ± 4.06 % ║
║ current code (memory) │      25 │ 248.25 op/sec │  ± 4.75 % ║
╟───────────────────────┼─────────┼───────────────┼───────────╢
║ Fastest test          │ Samples │        Result │ Tolerance ║
╟───────────────────────┼─────────┼───────────────┼───────────╢
║ old code (memory)     │      25 │ 265.48 op/sec │  ± 3.79 % ║
╚═══════════════════════╧═════════╧═══════════════╧═══════════╝
```

I don't see much of a difference (variance seems the same between new and old code, and old code is of course unchanged for this run), but the hashing algorithm is probably something _way_ more visible in micro benchmarks and not as part of an entire fake request flow (and again, my "benchmark" might be dumb)

lib/session.js
mcollina: https://www.npmjs.com/package/safe-stable-stringify < that's the one I use everywhere. Really recommended.

lib/session.js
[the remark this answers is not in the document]
mcollina: No worries, let's skip that optimization for now and we can investigate if it ever become a problem.

lib/session.js
SimenB: cool, will add!

mcollina:
```
+
+    // make 25 "requests" with the new session
+    for (let i = 0; i < 25; i++) {
+      await server.inject({ path: '/', headers: { cookie: `sessionId=${sessionId}` } })
+    }
+  }
```

This benchmark code is a bit flawed when using any I/O storage. You are essentially measuring Redis Round-Trip-Time. It's significantly better to:

1. run these in parallel, e.g. `Promise.all`
2. enable autoPipelining in ioredis.

I suspect that hashing would quickly become the bottleneck (as it did for etags).

lib/session.js
```diff
@@ -143,12 +151,34 @@ module.exports = class Session {
     return cookieSignature.sign(this.sessionId, this[secretKey])
   }
 
+  [hash] () {
+    const sess = this
+    const str = stringify(sess, function (key, val) {
```
SimenB, line 156: should we store the serialized session (`str` here) as well and use that in `save` instead of relying on whatever does the current serialization in the individual stores?

if yes, separate PR, but still

climba03003 (approved):
LGTM. with the hash algorithm update.

lib/session.js
SimenB: done

lib/session.js
mcollina: I wouldn't

mcollina (approved):
lgtm, this can land and we can change the hashing algorithm later

zekth (approved):
:v:

PR #71, opened 2022-03-23
fix: Unsign cookie, if already signed by cookie plugin

m0dch3n:
#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [] tests and/or benchmarks are included (I don't know how to test this properly, and I'm also missing the time...)
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #86, opened 2022-04-15
feat: add support for handling previously-signed cookies

### LINKED ISSUE #69: Session not working, when fastify-cookie plugin options use signed = true
### Prerequisites

### Fastify version

3.0.0

### Plugin version

6.4.0

### Node.js version

v16.13.2

### Operating system

macOS

12.3

### Description

Using `signed = true` in [fastify-cookie options](https://github.com/fastify/fastify-cookie#sending) causes that the sessionId is signed twice, because we pass the encryptedSessionId already... 

https://github.com/fastify/session/blob/73a4675b795e8ad3d020d5cd43de78374c09f8aa/lib/fastifySession.js#L126-L130

Once with the secret of `@fastify/session` and once with the secret of `fastify-cookie`

However on the decrypting side, we try to only decrypt the session encryption, but not the cookie encryption...
 https://github.com/fastify/session/blob/73a4675b795e8ad3d020d5cd43de78374c09f8aa/lib/fastifySession.js#L46
 
This causes that each time we generate a new session, because decryption fails with `false`

 
 

### Steps to Reproduce

Enable cookie signing on fastify-cookie plugin...

### Expected Behavior

_No response_

wilkmaia:
This PR increments the proposed solution in https://github.com/fastify/session/pull/71 by adding specific tests for the new flag and validating session is maintained throughout requests even if the cookie sent by the client had been previously signed.

Closes #69.

#### Checklist

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -121,16 +121,16 @@ function onRequest (options) {
     if (!sessionId) {
       newSession(secret, request, cookieOpts, idGenerator, done)
     } else {
+      let sessionToDecrypt = sessionId
+
       if (unsignSignedCookie) {
         const unsignedCookie = reply.unsignCookie(sessionId)
         if (unsignedCookie.valid) {
-          decryptSession(unsignedCookie.value, options, request, done)
-        } else {
-          newSession(secret, request, cookieOpts, idGenerator, done)
+          sessionToDecrypt = unsignedCookie.value
         }
-      } else {
-        decryptSession(sessionId, options, request, done)
       }
+
+      decryptSession(sessionToDecrypt, options, request, done)
     }
   }
 }
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -105,6 +105,7 @@ function decryptSession (sessionId, options, request, done) {
 }
 
 function onRequest (options) {
+  const unsignSignedCookie = options.unsignSignedCookie
   const cookieOpts = options.cookie
   const idGenerator = options.idGenerator
   return function handleSession (request, reply, done) {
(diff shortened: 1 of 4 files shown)


RafaelGSS (commented):
Can you add type tests for the new property as well?

wilkmaia:
Done, @RafaelGSS

RafaelGSS (approved):
LGTM.

mcollina (changes requested):
CI is failing for lack of coverage, could you add the missing tests?

simoneb:
@wilkmaia the build is broken, can you fix it?

test/session.test.js
```diff
@@ -758,3 +761,71 @@ test('only save session when it changes', async t => {
   // no set-cookie
   t.is(setCookieHeader2, undefined)
 })
+
+test('when unsignSignedCookie is true sessions should still be managed correctly', async t => {
+  t.plan(12)
```
simoneb, line 766: because you're not doing any assertions in callbacks that you need to wait for in order to consider the test complete, you can remove the .plan here

lib/fastifySession.js
```diff
@@ -120,7 +121,16 @@ function onRequest (options) {
...
       newSession(secret, request, cookieOpts, idGenerator, done)
     } else {
-      decryptSession(sessionId, options, request, done)
+      let sessionToDecrypt = sessionId
+
+      if (unsignSignedCookie) {
+        const unsignedCookie = reply.unsignCookie(sessionId)
+        if (unsignedCookie.valid) {
+          sessionToDecrypt = unsignedCookie.value
+        }
```
simoneb, line 130: what if the unsigned cookie is not valid? do you think it's sensible to ignore it and proceed?

simoneb (approved):

lib/fastifySession.js
wilkmaia: It's the same practical behavior we had before. What'll happen is that the `decryptSession` will be unable to decrypt the cookie and will, in turn, create a new session for the user. In case we can't unsign it I don't think there's much we can do but that. I've simply removed some bloat from this snippet.

Another approach that came to mind was to remove the `unsignSignedCookie` flag altogether and **always** check if the cookie is signed. In code it'd only remove the `if (unsignedCookie) {` check from around this block and we'd run the `reply.unsignCookie` on every request. Thinking of the library I think this would be a neat solution, given we'd be adding support for that feature transparently to the user, no need to tweak new flags. On the other hand, on my machine benchmarks were about 10% slower then (from ~273 ops/sec to ~254 ops/sec). Given that performance regression I decided to simply keep the approach that had been previously approved.

simoneb: thanks for the explanation! I'm happy to keep it as it is, just wanted to make sure we're intentional about it

PR #104, opened 2022-07-14
refactor: use optional chaining

Fdawgs:
Now that we are targeting Node 14 and above, we can take advantage of [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) and cut down the length of the conditions used in `if()` statements in this repo.

#### Checklist


lib/fastifySession.js
```diff
@@ -145,7 +145,7 @@ function onSend (options) {
 
     if (!shouldSaveSession(request, options.cookie, options.saveUninitialized)) {
       // if a session cookie is set, but has a different ID, clear it
-      if (request.cookies[options.cookieName] && request.cookies[options.cookieName] !== session.encryptedSessionId) {
+      if (request?.cookies[options.cookieName] !== session.encryptedSessionId) {
```
Eomm, line 148:
```suggestion
      if (request.cookies?.[options.cookieName] !== session.encryptedSessionId) {
```

mcollina (approved):
lgtm

(be careful that in v14 and possibly in v16 optional chaining is still slower than alternatives)

Fdawgs:
> lgtm
> 
> (be careful that in v14 and possibly in v16 optional chaining is still slower than alternatives)

Ah well I learned something new today. Not too sure it's worth doing this then if it affects performance.

Fdawgs:
@mcollina, as you're heavily involved in Node itself, any suggestions as to where I can learn more about performance differences and optimizations for Node?

mcollina:
> as you're heavily involved in Node itself, any suggestions as to where I can learn more about performance differences and optimizations for Node?

I don't think there is much documentation :(

Fdawgs:
> > as you're heavily involved in Node itself, any suggestions as to where I can learn more about performance differences and optimizations for Node?
> 
> 
> 
> I don't think there is much documentation :(

Ah okay, no worries, thanks anyway!

climba03003:
I think it failed because of recent change in `@fastify/cookie`?
https://github.com/fastify/fastify-cookie/commit/f41769472334276c97533fe0f662a4c94398e22b

mcollina:
Have you identified the reason @climba03003? Should we revert that change?

lib/fastifySession.js
```diff
@@ -145,7 +145,7 @@ function onSend (options) {
 
     if (!shouldSaveSession(request, options.cookie, options.saveUninitialized)) {
       // if a session cookie is set, but has a different ID, clear it
-      if (request.cookies[options.cookieName] && request.cookies[options.cookieName] !== session.encryptedSessionId) {
+      if (request.cookies?.[options.cookieName] !== session.encryptedSessionId) {
```
climba03003, line 148: Here should not be changed. Optional Chaining is not identical to what it is checking here.

```suggestion
      if (request.cookies[options.cookieName] && request.cookies[options.cookieName] !== session.encryptedSessionId) {
```

lib/fastifySession.js
```diff
@@ -63,7 +63,7 @@ function decryptSession (sessionId, options, request, done) {
         newSession(secret, request, cookieOpts, idGenerator, done)
         return
       }
-      if (session && session.expires && session.expires <= Date.now()) {
+      if (session?.expires <= Date.now()) {
```
climba03003, line 66:
```suggestion
      if (session?.expires && session.expires <= Date.now()) {
```

climba03003 (changes requested):

climba03003:
> Have you identified the reason @climba03003? Should we revert that change?

After debugging deeper, I spot the error is about the optional chaining.
Some changes is not identical to before.

climba03003 (approved):
CI is green. LGTM.

PR #114, opened 2022-08-10
Remove session id and encrypted session id from session data

rclmenezes:
Picking up where @simenb left off here: https://github.com/fastify/session/pull/101

- First, I rebased https://github.com/fastify/session/pull/101 on main. Needed to fix some Ava -> Tap stuff
- Next, I cleaned up the tests a little bit. There are some tests that modify a session or set expiration when that isn't the part of the test. It's best to keep tests more focused because otherwise bugs and slip through, as I'll talk about later.
- @SimenB had some [questions about the "multiple secrets" tests](https://github.com/fastify/session/pull/101#discussion_r909434094). The expiration was a red herring. The real issue was that sessionId/expiredSessionId wasn't in the session store, so the test always reset the sessionId. I made it cleaner.
- Finally, I found a bug in the original PR where saveUninitialized: true and maxAge wouldn't set a new cookie. It was because the hashing mechanism for dates converted all of them to the string "{}" by accident.

#### Checklist

#### Benchmarks

Running locally. Before:

```
$ node benchmark/bench.js
╔══════════════╤═════════╤═══════════════╤═══════════╤═════════════════════════╗
║ Slower tests │ Samples │        Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ file         │      25 │  32.68 op/sec │ ±  7.20 % │                         ║
║ redis        │      25 │ 139.43 op/sec │ ± 18.75 % │ + 326.64 %              ║
╟──────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ Fastest test │ Samples │        Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ memory       │      25 │ 760.16 op/sec │ ± 38.32 % │ + 2225.96 %             ║
╚══════════════╧═════════╧═══════════════╧═══════════╧═════════════════════════╝

✨  Done in 27.04s.
```

After:

```
$ node benchmark/bench.js
╔══════════════╤═════════╤═══════════════╤═══════════╤═════════════════════════╗
║ Slower tests │ Samples │        Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ file         │      25 │  27.58 op/sec │ ± 15.34 % │                         ║
║ redis        │      25 │  80.42 op/sec │ ± 25.62 % │ + 191.56 %              ║
╟──────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ Fastest test │ Samples │        Result │ Tolerance │ Difference with slowest ║
╟──────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ memory       │      25 │ 781.85 op/sec │ ± 37.90 % │ + 2734.57 %             ║
╚══════════════╧═════════╧═══════════════╧═══════════╧═════════════════════════╝

✨  Done in 31.01s.
```

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -62,7 +62,7 @@ function decryptSession (sessionId, options, request, done) {
         newSession(secret, request, cookieOpts, idGenerator, done)
         return
       }
-      if (session?.cookie?.expires && session.cookie.expires <= new Date()) {
+      if (session.cookie?.expires && session.cookie.expires <= Date.now()) {
         const restoredSession = Session.restore(
           request,
           idGenerator,
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -62,7 +62,7 @@ function decryptSession (sessionId, options, request, done) {
         newSession(secret, request, cookieOpts, idGenerator, done)
         return
       }
-      if (session?.expires && session.expires <= Date.now()) {
+      if (session?.cookie?.expires && session.cookie.expires <= Date.now()) {
         const restoredSession = Session.restore(
           request,
           idGenerator,
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -68,7 +68,8 @@ function decryptSession (sessionId, options, request, done) {
           idGenerator,
           cookieOpts,
           secret,
-          session
+          session,
+          decryptedSessionId
         )
 
         restoredSession.destroy(err => {
@@ -87,7 +88,8 @@ function decryptSession (sessionId, options, request, done) {
           idGenerator,
           cookieOpts,
           secret,
-          session
+          session,
+          decryptedSessionId
         )
       } else {
         request.session = Session.restore(
@@ -95,7 +97,8 @@ function decryptSession (sessionId, options, request, done) {
           idGenerator,
           cookieOpts,
           secret,
-          session
+          session,
+          decryptedSessionId
         )
       }
       done()
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -62,7 +62,7 @@ function decryptSession (sessionId, options, request, done) {
         newSession(secret, request, cookieOpts, idGenerator, done)
         return
       }
-      if (session?.cookie?.expires && session.cookie.expires <= Date.now()) {
+      if (session?.cookie?.expires && session.cookie.expires <= new Date()) {
         const restoredSession = Session.restore(
           request,
           idGenerator,
--- a/lib/session.js
+++ b/lib/session.js
@@ -22,7 +22,6 @@ const encryptedSessionIdKey = Symbol('encryptedSessionId')
 module.exports = class Session {
   constructor (request, idGenerator, cookieOpts, secret, prevSession = {}, sessionId = idGenerator(request)) {
     this[generateId] = idGenerator
-    this.expires = null
     this.cookie = new Cookie(cookieOpts)
     this[cookieOptsKey] = cookieOpts
     this[maxAge] = cookieOpts.maxAge
@@ -37,8 +36,7 @@ module.exports = class Session {
 
   touch () {
     if (this[maxAge]) {
-      this.expires = new Date(Date.now() + this[maxAge])
-      this.cookie.expires = this.expires
+      this.cookie.expires = new Date(Date.now() + this[maxAge])
     }
   }
 
@@ -70,7 +68,7 @@ module.exports = class Session {
 
   [addDataToSession] (prevSession) {
     for (const key in prevSession) {
-      if (!['expires', 'cookie', 'sessionId', 'encryptedSessionId'].includes(key)) {
+      if (!['cookie', 'sessionId', 'encryptedSessionId'].includes(key)) {
         this[key] = prevSession[key]
       }
     }
@@ -151,7 +149,8 @@ module.exports = class Session {
     const str = stringify(sess, function (key, val) {
       // ignore sess.cookie property
       if (this === sess && key === 'cookie') {
-        return
+        // we want `touch` to affect the hash of the session
+        return sess.cookie.expires
       }
 
       return val
@@ -178,9 +177,8 @@ module.exports = class Session {
   static restore (request, idGenerator, cookieOpts, secret, prevSession, sessionId) {
     const restoredSession = new Session(request, idGenerator, cookieOpts, secret, prevSession, sessionId)
     const restoredCookie = new Cookie(cookieOpts)
-    restoredCookie.expires = new Date(prevSession.expires)
+    restoredCookie.expires = new Date(prevSession.cookie.expires)
     restoredSession.cookie = restoredCookie
-    restoredSession.expires = restoredCookie.expires
     restoredSession[originalHash] = restoredSession[hash]()
     return restoredSession
   }
--- a/lib/session.js
+++ b/lib/session.js
@@ -70,7 +70,7 @@ module.exports = class Session {
 
   [addDataToSession] (prevSession) {
     for (const key in prevSession) {
-      if (!['expires', 'cookie'].includes(key)) {
+      if (!['expires', 'cookie', 'sessionId', 'encryptedSessionId'].includes(key)) {
         this[key] = prevSession[key]
       }
     }
--- a/lib/session.js
+++ b/lib/session.js
@@ -10,16 +10,17 @@ const stringify = configureStringifier({ bigint: false })
 
 const maxAge = Symbol('maxAge')
 const secretKey = Symbol('secretKey')
-const sign = Symbol('sign')
 const addDataToSession = Symbol('addDataToSession')
 const generateId = Symbol('generateId')
 const requestKey = Symbol('request')
 const cookieOptsKey = Symbol('cookieOpts')
 const originalHash = Symbol('originalHash')
 const hash = Symbol('hash')
+const sessionIdKey = Symbol('sessionId')
+const encryptedSessionIdKey = Symbol('encryptedSessionId')
 
 module.exports = class Session {
-  constructor (request, idGenerator, cookieOpts, secret, prevSession = {}) {
+  constructor (request, idGenerator, cookieOpts, secret, prevSession = {}, sessionId = idGenerator(request)) {
     this[generateId] = idGenerator
     this.expires = null
     this.cookie = new Cookie(cookieOpts)
@@ -29,10 +30,8 @@ module.exports = class Session {
     this[addDataToSession](prevSession)
     this[requestKey] = request
     this.touch()
-    if (!this.sessionId) {
-      this.sessionId = this[generateId](this[requestKey])
-      this.encryptedSessionId = this[sign]()
-    }
+    this[sessionIdKey] = sessionId
+    this[encryptedSessionIdKey] = cookieSignature.sign(this[sessionIdKey], secret)
     this[originalHash] = this[hash]()
   }
 
@@ -87,14 +86,14 @@ module.exports = class Session {
 
   destroy (callback) {
     if (callback) {
-      this[requestKey].sessionStore.destroy(this.sessionId, error => {
+      this[requestKey].sessionStore.destroy(this[sessionIdKey], error => {
         this[requestKey].session = null
 
         callback(error)
       })
     } else {
       return new Promise((resolve, reject) => {
-        this[requestKey].sessionStore.destroy(this.sessionId, error => {
+        this[requestKey].sessionStore.destroy(this[sessionIdKey], error => {
           this[requestKey].session = null
 
           if (error) {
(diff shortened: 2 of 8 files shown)


lib/fastifySession.js
```diff
@@ -62,13 +62,14 @@ function decryptSession (sessionId, options, request, done) {
         newSession(secret, request, cookieOpts, idGenerator, done)
         return
       }
-      if (session?.expires && session.expires <= Date.now()) {
+      if (session?.cookie?.expires && session.cookie.expires <= new Date()) {
         const restoredSession = Session.restore(
```
rclmenezes, line 66: We should re-evaluate:
- Whether or not to use object params. Once we get to like 8 it gets pretty confusing
- Whether Session.restore should exist at all. At this point, it's mostly a wrapper around prevSession

test/base.test.js
```diff
@@ -305,3 +316,25 @@ test('should not set session cookie if no data in session and saveUninitialized
   t.equal(response.statusCode, 200)
   t.ok(response.headers['set-cookie'] === undefined)
 })
+
+test('should set session cookie if saveUninitialized is false and data not modified but maxAge is on', async (t) => {
+  t.plan(2)
+  const options = {
+    cookie: {
+      maxAge: 42,
+      expires: new Date(Date.now() + 1000)
```
rclmenezes: TODO: I should remove this.

test/base.test.js
```diff
@@ -305,3 +316,25 @@ test('should not set session cookie if no data in session and saveUninitialized
   t.equal(response.statusCode, 200)
   t.ok(response.headers['set-cookie'] === undefined)
 })
+
+test('should set session cookie if saveUninitialized is false and data not modified but maxAge is on', async (t) => {
```
rclmenezes: Actually, this is probably the wrong behavior now that I think about it...

We want the cookie to get reset if the session already has data. If it has no data, we should leave it alone.

lib/session.js
```diff
@@ -147,16 +144,13 @@ module.exports = class Session {
...
-  }
-
   [hash] () {
     const sess = this
     const str = stringify(sess, function (key, val) {
       // ignore sess.cookie property
       if (this === sess && key === 'cookie') {
-        return
+        // we want `touch` to affect the hash of the session
+        return sess.cookie.expires?.toISOString()
```
rclmenezes, line 153: Otherwise, str might result in `expires: {}`

See:

```
> const { configure: configureStringifier } = require('safe-stable-stringify')
undefined
> const stringify = configureStringifier({ bigint: false })
undefined
> const str = stringify({
...   cookie: {
.....     expires: new Date()
.....   }
... }, function (key, val) {
...   // ignore sess.cookie property
...   if (key === 'cookie') {
.....     // we want `touch` to affect the hash of the session
.....     return val.expires
.....   }
... 
...   return val
... })
undefined
> console.log(str);
{"cookie":{}}
undefined
```

Which breaks some of the `saveUninitialized` stuff

test/base.test.js
```diff
@@ -194,12 +206,11 @@ test('should set new session cookie if expired', async (t) => {
   const plugin = fastifyPlugin(async (fastify, opts) => {
     fastify.addHook('onRequest', (request, reply, done) => {
       request.sessionStore.set('Qk_XT2K7-clT-x1tVvoY6tIQ83iP72KN', {
-        expires: Date.now() + 1000
+        cookie: { expires: new Date(Date.now() - 1000) }
```
rclmenezes, line 209: As Simon observed, this test was wrong?

test/base.test.js
```diff
@@ -278,7 +289,7 @@ test('should create new session if cookie contains invalid session', async (t) =
   const response = await fastify.inject({
     url: '/',
     headers: {
-      cookie: 'sessionId=Qk_XT2K7-clT-x1tVvoY6tIQ83iP72KN.B7fUDYXx9fXF9pNuL3qm4NVmSduLJ6kzCOPh5JhHGoE; Path=/; HttpOnly; Secure',
+      cookie: 'sessionId=Qk_XT2K7-clT-x1tVvoY6tIQ83iP72KN.ohnoinvalidnooooL3qm4NVmSduLJ6kzCOPh5JhHGoE; Path=/; HttpOnly; Secure',
```
rclmenezes, line 292: Made the difference a little more obvious :)

test/base.test.js
```diff
@@ -268,7 +279,7 @@ test('should create new session if cookie contains invalid session', async (t) =
   const plugin = fastifyPlugin(async (fastify, opts) => {
     fastify.addHook('onRequest', (request, reply, done) => {
       request.sessionStore.set('Qk_XT2K7-clT-x1tVvoY6tIQ83iP72KN', {
-        expires: Date.now() + 1000
+        test: {}
```
rclmenezes, line 282: This test has nothing to do with expiration, so I made the session contain something else. Also, expires lives in the cookie, not top-level

test/session.test.js
```diff
@@ -71,25 +73,6 @@ test('should add session.cookie object to request', async (t) => {
   t.equal(response.statusCode, 200)
 })
 
-test('should add session.expires object to request', async (t) => {
```
rclmenezes, line 74: Simon removed this test because it was redundant. I agree!

lib/session.js
```diff
@@ -10,36 +10,33 @@ const stringify = configureStringifier({ bigint: false })
...
     this[requestKey] = request
-    this.touch()
-    if (!this.sessionId) {
-      this.sessionId = this[generateId](this[requestKey])
-      this.encryptedSessionId = this[sign]()
-    }
+    this[sessionIdKey] = sessionId
+    this[encryptedSessionIdKey] = cookieSignature.sign(sessionId, secret)
     this[originalHash] = this[hash]()
+    this.touch() // Needs to happen after originalHash is set, in case maxAge forces an update
```
rclmenezes, line 34: This was a bug in the original PR. Added a test against it

test/base.test.js
```diff
@@ -64,16 +64,29 @@ test('should support multiple secrets', async (t) => {
...
-  t.equal(response.headers['set-cookie'].includes('aYb4uTIhdBXCfk_ylik4QN6-u26K0u0e'), false)
+  const responseForNewSecret = await fastify.inject({
+    url: '/',
+    headers: {
+      'x-forwarded-proto': 'https',
+      cookie: `sessionId=${sessionIdEncryptedWithNewSecret}; Path=/; HttpOnly; Secure`
+    }
+  })
+  t.equal(responseForNewSecret.statusCode, 200)
+  t.equal(responseForNewSecret.headers['set-cookie'].includes(sessionIdEncryptedWithNewSecret), true)
```
rclmenezes, line 89: Cleaning up this test!

test/session.test.js
```diff
@@ -309,7 +290,7 @@ test('should bubble up errors with destroy call if session expired', async (t) =
   const store = {
     set (id, data, cb) { cb(null) },
     get (id, cb) {
-      cb(null, { expires: Date.now() - 1000, cookie: { expires: Date.now() - 1000 } })
+      cb(null, { cookie: { expires: new Date(Date.now() - 1000) } })
```
rclmenezes, line 293: As I explain elsewhere, `expires` is supposed to be a Date, not a number

lib/session.js
Uzlopak: Should  we maybe serialize it to number by doing .getTime() instead of toISOString? Well i would prefer that.

lib/fastifySession.js
```diff
@@ -62,13 +62,14 @@ function decryptSession (sessionId, options, request, done) {
         newSession(secret, request, cookieOpts, idGenerator, done)
         return
       }
-      if (session?.expires && session.expires <= Date.now()) {
+      if (session?.cookie?.expires && session.cookie.expires <= new Date()) {
```
SimenB, line 65:
```suggestion
      if (session?.cookie?.expires && session.cookie.expires <= new Date().getTime()) {
```

I prefer being explicit 😀

lib/session.js
SimenB: nice!

SimenB (commented):
yay, thanks for picking this up!

lib/session.js
rclmenezes: For the purpose of hashing, that makes sense to me!

rclmenezes: Good call

rclmenezes:
```suggestion
        // we want `touch` to affect the hash of the session
        return sess.cookie.expires?.getTime?.()
```

Uzlopak: Why would you do the getTime?.

lib/fastifySession.js
rclmenezes:
```suggestion
      if (session?.cookie?.expires?.getTime && session.cookie.expires.getTime() <= new Date().getTime()) {
```

SimenB: I don't think the `getTime` of `session?.cookie?.expires?.getTime` is needed - if `expires` is defined so is its `getTime`. But beyond that I agree 👍

Uzlopak: you should prefer `Date.now()` over `new Date().getTime() `

But on the other hand,  you can compares Dates directly. So if session.cookie.expires is a Date you can compare it with Date.now()

Uzlopak:
```suggestion
      if (session.cookie?.expires && session.cookie.expires <= Date.now()) {
```

Uzlopak (changes requested):
I want to make a proper review. So dont merge it

Uzlopak:
@rclmenezes 

What is up with the failing unit test?

rclmenezes:
Interesting -- failing test works locally. It appears to be a race condition somewhere.

My guess is that `touch()` is running so quickly in CI that `session.isModified()` is returning true. One moment...

rclmenezes:
Ugh, I'm dumb. The cookie wasn't in `fastify.inject` so the test was wrong 🤦‍♂️ 

So it was making a new session. This would have a fresh `cookies.expires` value. Then .touch() would be called and either:
- .touch() would happen quickly enough where the fresh Date.now() was the same. Nothing modified so the cookie wouldn't change. This happened in CI.
- .touch() would take too long and the Date.now() was different, resulting in session.isModified() and triggering a set-cookie header. This happened locally and was what the test expected.

Uzlopak:
To be honest this PR is hard to digest. Can we please fix one issue at a time?

lib/session.js
rclmenezes: Oops, good point.

When some of the previous tests were passing numbers as `cookie.expires`, I was needing to do that. Now that they're fixed, we can avoid the `?.`

rclmenezes:
@Uzlopak happily! This PR is hard for me to digest too.

There were some tests that didn't do what they advertised, which caused a lot of grief for me. I can split it up into a PR that cleans up tests and a PR that does this change, if that helps?

lib/session.js
```diff
@@ -169,15 +163,28 @@ module.exports = class Session {
...
 
   isModified () {
-    return this[originalHash] !== this[hash]()
+    return Boolean(
+      this[originalHash] !== this[hash]() ||
+      // If maxAge is set, the session is modified.
+      // This is necessary because we don't read the cookie's expiration from the cookie itself.
+      // session.cookie.expires is initially set from Cookie(cookieOpts), so we cannot detect if it changed from maxAge alone.
+      this[maxAge]
+    )
```
SimenB, line 172:
```suggestion
    if (this[originalHash] !== this[hash]()) {
      return true
    }

    // If maxAge is set, the session is modified.
    // This is necessary because we don't read the cookie's expiration from the cookie itself.
    // session.cookie.expires is initially set from Cookie(cookieOpts), so we cannot detect if it changed from maxAge alone.
    return this[maxAge] != null
```

However, this seems wrong. Just setting a max age should never make `isModified` return `true` every single time

SimenB: >This is necessary because we don't read the cookie's expiration from the cookie itself.

We should probably do that, and isn't that what happens with your `return sess.cookie.expires?.getTime()` in the hash function?

rclmenezes: Ah -- I'm mistaken. This code is confusing :)

If `rolling` is false, Session.restore will be called, which addresses this.

Uzlopak:
Guys...  really split this PR in some logical parts. I am investigating some time on having some reliable benchmark as the current benchmark sucks currently. But yeah. Please lets split it.If you like we could first agree which are the parts and who wants to implement this or that feature/bug fix.

rclmenezes:
Okay I'm going to split this PR up into a bunch of parts:
1. I'm gonna clean up the tests. Add some constants, etc. Also I'll add the missing documentation for `rolling`.
2. `touch` should be implemented differently. First, it should not be called in the Session constructor; sessions should not be touched if `rolling` is false or `maxAge` is not set. I'll make `isTouched` a boolean property of `Session` and if it's true, the session will get saved (for now). Eventually, `if isTouched && !isModified` fastifySession can invoke `sessionStore.touch` to update TTL instead of `sessionStore.save`
3. I'll remove expires, sessionid and encryptedSessionid from the session state, like in the original PR.

Sound good?

Uzlopak:
Maybe simpler. would say following parts:
1. Add documentation for rolling
2. clean up tests, maybe another PR for for adding some constans
3. touch 
4. expires are Date
5. sessionId?
6. expiredSessionId?

Uzlopak:
Just get the PRs as small as possible keep them coming.

rclmenezes:
Sounds great.
1. Docs: https://github.com/fastify/session/pull/116
2. Test clean up: https://github.com/fastify/session/pull/117
3. A bugfix: https://github.com/fastify/session/pull/118

rclmenezes:
Closing in favor of:
https://github.com/fastify/session/pull/121

PR #125, opened 2022-08-15
feat: @fastify/cookie responsible for the signing and unsigning

Uzlopak:
This PR couples @fastify/cookie stronger to @fastify/session. So instead of duplicating signing and unsigning logic in @fastify/session, we expect that setting the signing related stuff in @fastify/cookie and we just consume the logic in @fastify/session.

@rclmenezes 
I think before we merge other patches, we should first merge this.

To review this PR i think, it is easier by going through the commits. I could not break it down into much smaller PRs as it is basically one big change.

#### Checklist


test/session.test.js
```diff
@@ -105,7 +104,7 @@ test('should allow get/set methods for fetching/updating session values', async
 test('should use custom sessionId generator if available (without request)', async (t) => {
   t.plan(2)
   const fastify = await buildFastify((request, reply) => {
-    t.equal(request.session.sessionId.startsWith('custom-'), false)
+    t.equal(request.session.sessionId.startsWith('custom-'), true)
```
Uzlopak, line 107: This is the only unit test i did change. I actually think that this is the correct behaviour.

Uzlopak: it seems that there was a race condition in the unit tests

climba03003 (changes requested):
By comparing `express-session`, the `secret` option should stay in here.

Uzlopak:
I think my approach is much cleaner, as we just need to setup fastify-cookie correct and then can use the actual methods instead of taking also care of the secrect management.

Uzlopak:
I think my approach is much cleaner, as we just need to setup fastify-cookie correct and then can use the actual methods instead of taking also care of the secrect management.

Uzlopak:
I think my approach is much cleaner, as we just need to setup fastify-cookie correct and then can use the actual methods instead of taking also care of the secrect management.

climba03003:
> I think my approach is much cleaner, as we just need to setup fastify-cookie correct and then can use the actual methods instead of taking also care of the secrect management.

It require the user to be clear that they shouldn't forget the `secret` in registering `@fastify-cookie`.
There is no way to block them and remind them to add one before application startup.

It is totally different then way before.

mcollina (changes requested):
Here is a few questions:

1. What happens if there is no secret set to both `@fastify/cookie` and `@fastify/session`? Let's make sure an error is thrown at some point.

2. Make sure to throw if `secret` is passed to `@fastify/session`.

package.json
```diff
@@ -19,7 +19,7 @@
   "author": "Denis Fäcke",
   "license": "MIT",
   "dependencies": {
-    "cookie-signature": "^1.1.0",
+    "@fastify/cookie": "^7.4.0",
```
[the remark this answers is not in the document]
SimenB: can `fastify-plugin` ensure the version of `@fastify/cookie` is at least 7.4? or is that out of scope for that module, and it might fail at runtime when accessing non-existent methods?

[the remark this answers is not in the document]
mcollina: You can require for a decorator to exist: https://github.com/fastify/fastify-plugin#dependencies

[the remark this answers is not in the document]
SimenB: Ah, right! should add that 👍

Uzlopak:
@mcollina 
Created https://github.com/fastify/fastify-cookie/pull/203

So when no secret is set, we can feature detect if the decorators exist. If they dont we can throw a secret not set error. Obviously I can only write corresponding unit tests if the PR  is merged and a new release of @fastify/cookie was published.

mcollina (approved):
lgtm

mcollina:
@Uzlopak please wait for @climba03003 and @SimenB

Uzlopak:
will do

Uzlopak:
@mcollina 

I still have to add that the plugin throws an error if a secret was provided.

climba03003:
It do not convince me to provide `secret` in `@fastify/cookie`.

1. It leaks of length check, user can use weak `secret` which can't in currently.
2. User can not switch from `express-session` to `@fastify/session` with just config copy and paste.

Point `2.` is not as important as `1.`.
If it really concern about code de-duplication, why can't we use the `signer factory` from `@fastify/cookie`?

lib/session.js
```diff
@@ -36,7 +32,7 @@ module.exports = class Session {
     this.touch()
     if (!this.sessionId) {
       this.sessionId = this[generateId](this[requestKey])
-      this.encryptedSessionId = this[sign]()
+      this.encryptedSessionId = this[requestKey].signCookie(this.sessionId)
```
SimenB, line 35:
```suggestion
      this.encryptedSessionId = request.signCookie(this.sessionId)
```

package.json
```diff
@@ -19,7 +19,7 @@
   "author": "Denis Fäcke",
   "license": "MIT",
   "dependencies": {
-    "cookie-signature": "^1.1.0",
+    "@fastify/cookie": "^8.0.0",
```
SimenB, line 22: should be in dev deps

SimenB (approved):

rclmenezes:
My last few patches have been all about making @fastify/session compatible with express-session.

This PR creates a new incompatibility: what if someone provides a secret to both `express-cookie` and `express-session`? They probably _shouldn't_, but the effect would be that non-session-related cookies have one secret and sessions have another.

Uzlopak:
Dont worry. Working on a different branch.

Uzlopak:
Closing in favor of #129

PR #129, opened 2022-08-16
feat: use @fastify/cookie-signer-logic for signing and unsigning

Uzlopak:
This is an alternative implementation of #125 

#### Checklist


Uzlopak:
@mcollina 
@simoneb 
@climba03003 
@rclmenezes

This is an alternative implementation of #125 . It should be better compatible with express-session and takes the remarks of @climba03003  into account.

lib/fastifySession.js
```diff
@@ -4,16 +4,22 @@ const fp = require('fastify-plugin')
 const idGenerator = require('./idGenerator')()
 const Store = require('./store')
 const Session = require('./session')
-const cookieSignature = require('cookie-signature')
+let Signer
 
 function session (fastify, options, next) {
+  if (!Signer) {
+    Signer = require('@fastify/cookie').Signer
```
SimenB, line 11: Why is this lazy?

Uzlopak: Because @fastify/cookie is a dev dependency. If i load it directly then it directly throws an error and never reaches the dependency check of fastify-plugin

SimenB: Aha, gotcha. And `Signer` should not be decorated on `fastify`?

Uzlopak: feels kind of wrong to decorate Signer on fastify

climba03003 (approved):
LGTM. It is definitely better.

PR #132, opened 2022-08-17
fix unit test

Uzlopak:
I really dont know how to describe this issue. I dont know why I need to pass a new Object for register. If I just pass sessionOptions directly then the fixed unit test will fail. Super strange, and makes me wonder if there is any documentation regarding this odd behaviour in fastify core?

#### Checklist

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -187,17 +187,18 @@ function session (fastify, options, next) {
   }
 
   function ensureDefaults (options) {
-    options.store = options.store || new Store()
-    options.idGenerator = options.idGenerator || idGenerator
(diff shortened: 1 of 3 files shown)


mcollina:
The tests are not failing on master.

Uzlopak:
@mcollina 

I know they are not failing, but the 'should use custom sessionId generator if available (without request)' test should fail on master. If you comment out all tests before that test, it will fail on master.

mcollina:
I think the problem is that this module is modifying the options object instead of creating a copy of it. Could you make it so?

Uzlopak:
@mcollina 
I made ensureDefaults immutable

mcollina (approved):
lgtm

PR #137, opened 2022-08-24
use redis pipeline

Uzlopak:
#### Checklist

### DIFF
--- a/test/redis.test.js
+++ b/test/redis.test.js
@@ -29,6 +29,5 @@ if (process.env.CI) {
   mock('./base.test', { '../lib/store': RedisStore })
   mock('./cookie.test', { '../lib/store': RedisStore })
   mock('./session.test', { '../lib/store': RedisStore })
-  mock('./idGenerator.test', { '../lib/store': RedisStore })
   mock('./store.test', { '../lib/store': RedisStore })
 }
--- a/test/redis.test.js
+++ b/test/redis.test.js
@@ -1,6 +1,6 @@
 'use strict'
 
-if (process.env.CI) {
+if (!process.env.NO_REDIS) {
   const { mock, before, afterEach, teardown } = require('tap')
   const redisStoreFactory = require('connect-redis')
   const Redis = require('ioredis')
--- a/test/redis.test.js
+++ b/test/redis.test.js
@@ -1,7 +1,6 @@
 'use strict'
 
 if (process.env.CI) {
-
   const { mock, before, afterEach, teardown } = require('tap')
   const redisStoreFactory = require('connect-redis')
   const Redis = require('ioredis')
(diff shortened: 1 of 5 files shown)


Uzlopak:
I am tired. Will work tomorrow on this.

test/redis.test.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+if (process.env.CI) {
```
mcollina, line 3: why have you introduced this `if` here? If we are running tests for Redis, let's run them locally too.

test/redis.test.js
```diff
@@ -0,0 +1,33 @@
...
+  afterEach(async () => {
+    await client.flushall()
+  })
+
+  teardown(() => client.disconnect())
+
+  mock('./base.test', { '../lib/store': RedisStore })
+  mock('./cookie.test', { '../lib/store': RedisStore })
+  mock('./session.test', { '../lib/store': RedisStore })
+  mock('./store.test', { '../lib/store': RedisStore })
```
mcollina, line 32: As I understand this code, you'd run the same tests but mock the full `lib/store` module. I would rather prefer an approach that did not rely on mocking - I fear this is going to be very brittle to maintain over time.

mcollina (changes requested):

test/redis.test.js
Uzlopak: Well it I use here mock to make it behave like proxyquire. Instead of loading the MemoryStore we actually load the real RedisStore. Does this make sense to you?

test/redis.test.js
Uzlopak: You can run unit tests with redis by doing npm run test:ci 

I just dont want to force a dev, who is using windows, to go to the hassle to activate WSL2 to be able to contribute. 

Also we have alot of flushall calls, so it has side effects. I dont want that somebody later complains, that by running the unit tests in his docker container, which is in his test or prod environment, and having a redis bound to localhost to be accidently flushed, because he tested the unit tests shipped with the npm-module.

Uzlopak: If you insist, then I will change. Then I would prefer to introduce a NO_REDIS env variable to disable the unit tests, if needed.

Uzlopak: created a npm script test:unit:memory with NO_REDIS env flag.

test/redis.test.js
mcollina: Indeed, I don't think this would be easy to maintain long-term.

Uzlopak: I close this for now, as I need the changes in the actual codebase to also add file store tests. 

But I would like to know, why this simple change is not easy to maintain long-term.

mcollina: It's unscalable to the addition of more databases as running multiple dbs in the same action is a recipe for flaky tests.
It's incredibly hard to debug because it runs all the tests at the same time too.

A better approach is to set a config/db as an environment variable and kick off the full test suite for every database.

PR #138, opened 2022-08-24
Revert "refactor store to es6 class"

Uzlopak:
Reverts fastify/session#122

If it is a ES6 class it will have regressions, as some store implementations dont extend the store class but call the function instead. This results that the benchmarks failed

see https://github.com/valery-barysok/session-file-store/blob/b89a42b00f24e720bb9d07cc7ddf72ef6c43c7bc/lib/session-file-store.js#L26

### DIFF
--- a/lib/store.js
+++ b/lib/store.js
@@ -1,25 +1,28 @@
 'use strict'
 
-const { EventEmitter } = require('events')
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #141, opened 2022-08-24
dont use storeMap to check for store length

Uzlopak:
This is a deviation of #137 
it makes it possible to also check file store and redis store when we agree on how to test them

new is length method of store, which is also provided by file store and redis store. 

#### Checklist

### DIFF
--- a/lib/store.js
+++ b/lib/store.js
@@ -25,8 +25,4 @@ Store.prototype.destroy = function destroy (sessionId, callback) {
   callback()
 }
 
-Store.prototype.length = function length (callback) {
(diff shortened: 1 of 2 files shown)


lib/store.js
```diff
@@ -25,4 +25,8 @@ Store.prototype.destroy = function destroy (sessionId, callback) {
   callback()
 }
 
+Store.prototype.length = function length (callback) {
+  callback(null, this.store.size)
+}
```
mcollina, line 30: I don't think we want to add more methods to store.

PR #118, opened 2022-08-12
Bugfix: when there are secrets, we should re-sign with latest

rclmenezes:
This PR is dependent on:
https://github.com/fastify/session/pull/120

Only the last commit is relevant for review for now.

Turns out, we were re-signing cookies with the new secret.

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #143, opened 2022-08-24
prevent encryptedSessionId from being stored in the sessionStore

Uzlopak:
#### Checklist


Uzlopak:
@rclmenezes

lib/session.js
```diff
@@ -24,6 +25,7 @@ module.exports = class Session {
     this[cookieSignerKey] = cookieSigner
     this[requestKey] = request
     this[sessionIdKey] = sessionId
+    this[encryptedSessionIdKey] = (prevSession && prevSession[sessionIdKey] === sessionId && prevSession[encryptedSessionIdKey]) || cookieSigner.sign(this.sessionId)
```
Uzlopak, line 28: @rclmenezes 
I deviated from your PR. If the prevSession has sessionIdKey and sessionId is the same, then we dont need to encrypt again, as it should be the same. Do you agree?

rclmenezes: Depends. If there are multiple secrets, I think anything with the old secret should be resigned with the new secret. Hence this PR:
https://github.com/fastify/session/pull/118

But honestly, I don't feel very strongly about it 🤷‍♂️

Uzlopak: For multiple unit tests we have this unit test

https://github.com/fastify/session/blob/b7661cdc46c228547d36fe6cc50a8e1190131b13/test/base.test.js#L79

So we have that case covered.

rclmenezes: Unfortunately, this test doesn't properly cover that case.

Adapt the test from my original PR and we'll see a failure:

```typescript
test('should support multiple secrets', async (t) => {
  t.plan(4)
  const sign = require('@fastify/cookie/signer').sign
  const newSecret = 'geheim'
  const options = {
    secret: [newSecret, DEFAULT_SECRET]
  }

  const sessionId = 'aYb4uTIhdBXCfk_ylik4QN6-u26K0u0e'
  const sessionIdSignedWithOldSecret = sign(sessionId, DEFAULT_SECRET)
  const sessionIdSignedWithNewSecret = sign(sessionId, newSecret)
  const htmlEncodedSessionIdSignedWithNewSecret = 'aYb4uTIhdBXCfk_ylik4QN6-u26K0u0e.InCp31AuDa7DX%2F8rGBz8RMFiCpmUtjcF%2BS7Aco7tur8'

  const plugin = fastifyPlugin(async (fastify, opts) => {
    fastify.addHook('onRequest', (request, reply, done) => {
      request.sessionStore.set(sessionId, {
        sessionId,
        encryptedSessionId: sessionIdSignedWithOldSecret
      }, done)
    })
  })
  function handler (request, reply) {
    request.session.test = {}
    reply.send(200)
  }
  const fastify = await buildFastify(handler, options, plugin)
  t.teardown(() => fastify.close())

  const responseForOldSecret = await fastify.inject({
    url: '/',
    headers: {
      'x-forwarded-proto': 'https',
      cookie: `sessionId=${sessionIdSignedWithOldSecret}; Path=/; HttpOnly; Secure`
    }
  })

  t.equal(responseForOldSecret.statusCode, 200)
  // It will be replaced with the new secret!
  t.equal(responseForOldSecret.headers['set-cookie'].includes(htmlEncodedSessionIdSignedWithNewSecret), true)

  const responseForNewSecret = await fastify.inject({
    url: '/',
    headers: {
      'x-forwarded-proto': 'https',
      cookie: `sessionId=${sessionIdSignedWithNewSecret}; Path=/; HttpOnly; Secure`
    }
  })
  t.equal(responseForNewSecret.statusCode, 200)
  t.equal(responseForNewSecret.headers['set-cookie'].includes(htmlEncodedSessionIdSignedWithNewSecret), true)
})
```

Why? What's the difference between this and the current test?

The current test does not set `encryptedSessionId` in the session prior to running the test. So `encryptedSessionId` never gets copied over here:
https://github.com/fastify/session/blob/master/lib/session.js#L31-L33

Since `encryptedSessionId` isn't set, it gets a new signature.

In production, `encryptedSessionId` would be in the previous session.

rclmenezes: I was running on main, not on this branch.

On this branch, we would need to remove `sessionId` and `encryptedSessionId` from the store. Otherwise we'll see the setter error.

But -- it does pass on your branch already so we're all set :)

Uzlopak: so you would approve this PR? Or am I missing something?

rclmenezes: Sorry sorry it's been a pretty terrible day. Reviewing now.

rclmenezes (approved):
LGTM!

Uzlopak:
Thank you, this gives me more confidence :)

I added now some other reviewers from the team to get also their approval.

lib/session.js
SimenB, line 28: this logic is hard to read, could you split it into some `if`s instead of 3 inline conditionals?

test/base.test.js
```diff
@@ -36,9 +36,8 @@ test('should save the session properly', async (t) => {
       const keys = Object.keys(session)
 
       // Only storing three keys: cookie, encryptedSessionId and test
-      t.equal(keys.length, 3)
+      t.equal(keys.length, 2)
       t.ok(session.cookie)
-      t.ok(session.encryptedSessionId)
```
SimenB, line 41: this is still here, only via the getter, no? so `keys.length` is decreased, but this assertion would still work

SimenB (changes requested):

test/base.test.js
Uzlopak: You are right.. lol. We need to check if it is not in the store saved

lib/session.js
Uzlopak: but it is branchless code?!

test/base.test.js
Uzlopak: created a TestStore, which uses JSON.stringify and parse to throw away the symbol keys.

lib/session.js
Uzlopak: I reformatted it. Is it acceptable now :)?

test/base.test.js
```diff
@@ -35,11 +35,17 @@ test('should save the session properly', async (t) => {
       const session = [...storeMap.entries()][0][1]
       const keys = Object.keys(session)
 
-      // Only storing three keys: cookie, encryptedSessionId and test
-      t.equal(keys.length, 3)
+      // Only storing three keys: cookie and test
```
SimenB, line 38:
```suggestion
      // Only storing two keys: cookie and test
```

lib/session.js
```diff
@@ -24,6 +25,11 @@ module.exports = class Session {
     this[cookieSignerKey] = cookieSigner
     this[requestKey] = request
     this[sessionIdKey] = sessionId
+    this[encryptedSessionIdKey] = (
+      prevSession &&
+      prevSession[sessionIdKey] === sessionId &&
+      prevSession[encryptedSessionIdKey]
+    ) || cookieSigner.sign(this.sessionId)
```
SimenB, line 32:
```suggestion
    this[encryptedSessionIdKey] =
      prevSession && prevSession[sessionIdKey] === sessionId
        ? prevSession[encryptedSessionIdKey]
        : cookieSigner.sign(this.sessionId);
```

right?

test/base.test.js
```diff
@@ -35,11 +35,17 @@ test('should save the session properly', async (t) => {
...
+      t.equal(keys.length, 2)
+      t.ok(keys.indexOf('cookie') !== -1)
+      t.ok(keys.indexOf('test') !== -1)
+      t.ok(keys.indexOf('sessionId') === -1)
+      t.ok(keys.indexOf('encryptedSessionId') === -1)
```
SimenB, line 43: `includes` instead of `indexOf`?

lib/session.js
Uzlopak: Ternary operator is branching. It should be theoretically slower. I will make benchmarks to see if it has some impact.

test/base.test.js
Uzlopak: Only if you provide the suggestions :)

Uzlopak: done

lib/session.js
Uzlopak: The ternary operator makes it 5 % slower.

test/TestStore.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+const { EventEmitter } = require('events')
+
+class TestStore extends EventEmitter {
```
SimenB, line 5: why do we need this? Looks identical to https://github.com/fastify/session/blob/master/lib/store.js except being a `class`

SimenB (approved):
👍

test/TestStore.js
Uzlopak: simplified it. it differs from the .set()-method.

Uzlopak: Also I dont want that somebodys code breaks just because I modified the Store for the tests. Imagine somebody really using the memorystore for some reason and somehow stores circular references or so. Also it makes it possible to later modify the store to be a RedisStore or so.

SimenB: nice, thanks for clarifying!

Issue #147, opened 2022-08-24
Make sure we set session cookie if saveUninitialized is false and maxAge is on


rclmenezes:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4

### Plugin version

9

### Node.js version

16

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

*

### Description

Currently, if maxAge is on and saveUninitialized is false, we don't re-save the session.

See this for original discussion:
https://github.com/fastify/session/pull/144#issuecomment-1225872415

Fix is putting the cookie's expiration in the hash:

```typescript
if (this === sess && key === 'cookie') {
        // we want `touch` to affect the hash of the session
        return sess.cookie.expires?.getTime()
      }
}
```

### Steps to Reproduce

```typescript
test('should set session cookie if saveUninitialized is false and maxAge is on', async (t) => {
  t.plan(2)
  const options = {
    cookie: {
      maxAge: 42
    },
    secret: 'cNaoPYAwF60HZJzkcNaoPYAwF60HZJzk',
    saveUninitialized: false
  }
  const plugin = fastifyPlugin(async (fastify, opts) => {
    fastify.addHook('onRequest', (request, reply, done) => {
      request.sessionStore.set(DEFAULT_SESSION_ID, {
        // In this scenario, maxAge would have set expires in a previous request
        cookie: {
          expires: new Date(Date.now() + 1000)
        }
      }, done)
    })
  })
  const fastify = await buildFastify((request, reply) => reply.send(200), options, plugin)
  t.teardown(() => fastify.close())

  const response = await fastify.inject({
    url: '/',
    headers: { cookie: DEFAULT_COOKIE, 'x-forwarded-proto': 'https' }
  })

  t.equal(response.statusCode, 200)
  t.ok(response.headers['set-cookie'])
})
```

### Expected Behavior

The session should re-save

PR #149, opened 2022-08-24
Resave session cookie if saveUninitialized is false and rolling is true

rclmenezes:
Edit: I'm breaking up this PR, setting back to draft.

- To do this, we need to make sure that cookie.expires is used in the cookie hash
- This in turn, required that expires be an actual Date in our tests

See issue: https://github.com/fastify/session/issues/147

#### Checklist

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -39,7 +39,7 @@ function getExpires (cookieOpts) {
   if (cookieOpts.maxAge) {
     return new Date(Date.now() + cookieOpts.maxAge)
   } else if (cookieOpts.expires) {
-    return cookieOpts.expires
+    return new Date(cookieOpts.expires)
   }
   return null
 }
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -53,21 +53,24 @@ function session (fastify, options, next) {
         }
         return
       }
+
       if (!session) {
         request.session = new Session(request, idGenerator, cookieOpts, cookieSigner)
         done()
         return
       }
-      if (session.cookie?.expires && session.cookie.expires <= Date.now()) {
-        const restoredSession = new Session(
-          request,
-          idGenerator,
-          cookieOpts,
-          cookieSigner,
-          session,
-          decryptedSessionId
-        )
 
+      request.session = new Session(
+        request,
+        idGenerator,
+        cookieOpts,
+        cookieSigner,
+        session,
+        decryptedSessionId
+      )
+
+      if (request.session.cookie.expires && request.session.cookie.expires.getTime() <= Date.now()) {
+        const restoredSession = request.session
         restoredSession.destroy(err => {
           if (err) {
             done(err)
@@ -78,14 +81,7 @@ function session (fastify, options, next) {
         })
         return
       }
-      request.session = new Session(
-        request,
-        idGenerator,
-        cookieOpts,
-        cookieSigner,
-        session,
-        decryptedSessionId
-      )
+
       done()
     })
   }
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -152,7 +152,7 @@ function session (fastify, options, next) {
         }
         reply.setCookie(
           cookieName,
-          `${cookiePrefix}${session.encryptedSessionId}`,
+          hasCookiePrefix ? `${cookiePrefix}${session.encryptedSessionId}` : session.encryptedSessionId,
           session.cookie.options(isConnectionSecure(request))
         )
         done()
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -64,7 +64,7 @@ function session (fastify, options, next) {
         return
       }
 
-      request.session = new Session(
+      const restoredSession = new Session(
         request,
         idGenerator,
         cookieOpts,
@@ -73,8 +73,7 @@ function session (fastify, options, next) {
         decryptedSessionId
       )
 
-      if (request.session.cookie.expires && request.session.cookie.expires.getTime() <= Date.now()) {
-        const restoredSession = request.session
+      if (restoredSession.cookie.expires && restoredSession.cookie.expires.getTime() <= Date.now()) {
         restoredSession.destroy(err => {
           if (err) {
             done(err)
@@ -86,6 +85,7 @@ function session (fastify, options, next) {
         return
       }
 
+      request.session = restoredSession
       done()
     })
   }
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -123,7 +123,8 @@ function session (fastify, options, next) {
 
   function onSend (options) {
     const cookieOpts = options.cookie
-    const saveUninitialized = options.saveUninitialized
+    const saveUninitializedSession = options.saveUninitialized
+    const rollingSessions = options.rolling
 
     return function saveSession (request, reply, payload, done) {
       const session = request.session
@@ -133,7 +134,9 @@ function session (fastify, options, next) {
       }
 
       const cookieSessionId = getCookieSessionId(request)
-      if (!shouldSaveSession(request, cookieOpts, saveUninitialized)) {
+      const shouldSaveSession = getShouldSaveSession(request, cookieSessionId, saveUninitializedSession, rollingSessions)
+      const isInsecureConnection = cookieOpts.secure === true && !isConnectionSecure(request)
+      if (!shouldSaveSession || isInsecureConnection) {
         // if a session cookie is set, but has a different ID, clear it
         if (cookieSessionId && cookieSessionId !== session.encryptedSessionId) {
           reply.clearCookie(cookieName)
@@ -188,19 +191,14 @@ function session (fastify, options, next) {
 
   function isConnectionSecure (request) {
     return (
-      request.raw.socket?.encrypted === true ||
-    request.headers['x-forwarded-proto'] === 'https'
+      request.raw.socket?.encrypted === true || request.headers['x-forwarded-proto'] === 'https'
     )
   }
 
-  function shouldSaveSession (request, cookieOpts, saveUninitialized) {
-    if (!saveUninitialized && !request.session.isModified()) {
-      return false
-    }
-    if (cookieOpts.secure !== true || cookieOpts.secure === 'auto') {
-      return true
-    }
-    return isConnectionSecure(request)
+  function getShouldSaveSession (request, cookieId, saveUninitializedSession, rollingSessions) {
+    return cookieId !== request.session.encryptedSessionId
+      ? saveUninitializedSession || request.session.isModified()
+      : rollingSessions || (Boolean(request.session.cookie.expires) && request.session.isModified())
   }
 
   function option (options, key, def) {
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -90,13 +90,18 @@ function session (fastify, options, next) {
     })
   }
 
-  function getCookieSessionId (request) {
-    const cookieValue = request.cookies[cookieName]
-    if (cookieValue && hasCookiePrefix && cookieValue.startsWith(cookiePrefix)) {
-      return cookieValue.slice(cookiePrefixLength)
+  const getCookieSessionId = hasCookiePrefix
+    ? function getCookieSessionId (request) {
+      const cookieValue = request.cookies[cookieName]
+      return (
+        cookieValue &&
+        cookieValue.startsWith(cookiePrefix) &&
+        cookieValue.slice(cookiePrefixLength)
+      )
     }
-    return cookieValue
-  }
+    : function getCookieSessionId (request) {
+      return request.cookies[cookieName]
+    }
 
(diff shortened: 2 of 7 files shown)


test/base.test.js
```diff
@@ -455,3 +455,41 @@ test('should handle custom signer', async (t) => {
...
+    Date.now = DateNow
+  })
+
+  const response = await fastify.inject({
+    url: '/',
+    headers: { cookie: DEFAULT_COOKIE, 'x-forwarded-proto': 'https' }
+  })
+
+  t.equal(response.statusCode, 200)
+  t.match(response.headers['set-cookie'], RegExp(`sessionId=.*; Path=/; Expires=${new Date(now + 42).toUTCString()}; HttpOnly`))
```
Eomm, line 494: Why is it `+42` instead of `+10`? Should the `sessionStore.set` take priority?

test/base.test.js
```diff
@@ -455,3 +455,41 @@ test('should handle custom signer', async (t) => {
...
+    cookie: {
+      maxAge: 42
+    },
+    secret: DEFAULT_SECRET,
+    saveUninitialized: false
+  }
+  const plugin = fastifyPlugin(async (fastify, opts) => {
+    fastify.addHook('onRequest', (request, reply, done) => {
+      request.sessionStore.set(DEFAULT_SESSION_ID, {
+        // In this scenario, maxAge would have set expired in a previous request
```
Eomm, line 475: Does it worth testing it in a dedicated test case?

lib/session.js
```diff
@@ -33,11 +33,11 @@ module.exports = class Session {
       }
     }
 
-    this.touch()
     if (!this.encryptedSessionId) {
       this.encryptedSessionId = cookieSigner.sign(this.sessionId)
     }
     this[originalHash] = this[hash]()
+    this.touch()
```
Uzlopak, line 40: Do we need .touch() when we merge #152 ? Or did I make a mistake in #152

test/base.test.js
```diff
@@ -461,3 +461,41 @@ test('should handle custom signer', async (t) => {
   t.equal(response.headers['set-cookie'].includes(DEFAULT_ENCRYPTED_SESSION_ID), false)
   t.match(response.headers['set-cookie'], /sessionId=[\w-]{32}.[\w-%]{43,135}; Path=\/; HttpOnly; Secure/)
 })
+
+test('should set session cookie if saveUninitialized is false and maxAge is on', async (t) => {
```
Uzlopak, line 465: This unit test is contradicting line 383 test: "should not set session cookie if no data in session and saveUninitialized is false"

test/base.test.js
rclmenezes: Nope, this hook runs before `@fastify/session`. This is the pattern we use in most of our tests to simulate a pre-existing session.

test/base.test.js
rclmenezes: We do here already:
https://github.com/fastify/session/blob/847826c6409931e107598fbf5096aac047603d3b/test/cookie.test.js#L98

But you're right, doing two requests in this test would make it less brittle.

lib/session.js
rclmenezes: Ah crap good catch.

I didn't merge your code in, let me do that...

test/base.test.js
rclmenezes: There's a key difference in these tests though! In this test, the session is already initialized. 

`saveUninitialized: false` should only prevent uninitialized sessions from being saved. Existing sessions should still be saved

mcollina:
CI is failing

Uzlopak:
If you have the overview, you can split the mighty test file to fastifySession.rolling.test.js and fastifySession.saveUnitialized.test.js etc. So that we have each aspect/option in a separate file.

rclmenezes:
Hi @Uzlopak -- 

I broke this PR up into three pretty legible commits. Let me know if you want them to be separate PRs.

The first commit cleans up our dates. `cookie.expires` should always be a Date. 
The second commit cleans up cookiePrefix a little bit.
The third commit implements rolling.

Lemme know what you think!

lib/cookie.js
```diff
@@ -39,7 +39,7 @@ function getExpires (cookieOpts) {
   if (cookieOpts.maxAge) {
     return new Date(Date.now() + cookieOpts.maxAge)
   } else if (cookieOpts.expires) {
-    return cookieOpts.expires
+    return new Date(cookieOpts.expires)
```
rclmenezes, line 42: `expires` should _always_ be a Date. The conversion always happens here.

lib/fastifySession.js
```diff
@@ -53,21 +57,24 @@ function session (fastify, options, next) {
...
-        )
 
+      request.session = new Session(
+        request,
+        idGenerator,
+        cookieOpts,
+        cookieSigner,
+        session,
+        decryptedSessionId
+      )
```
rclmenezes, line 74: We might as well push this up.

It's pretty helpful because it converts `cookie.expires` into a date for us, so we can use `getTime()` for comparison against `Date.now()`.

test/session.test.js
```diff
@@ -798,6 +799,56 @@ test('only save session when it changes', async t => {
   t.equal(setCookieHeader2, undefined)
 })
 
+test('when rolling is true, keep saving the session', async t => {
```
rclmenezes, line 802: If the session already exists, we need to save the session again when `rolling` is set, regardless of whether or not the session was modified.

This replaces my previous code in this PR, when I used `maxAge` as a crutch for this.

lib/fastifySession.js
```diff
@@ -192,19 +191,14 @@ function session (fastify, options, next) {
...
-      return false
-    }
-    if (cookieOpts.secure !== true || cookieOpts.secure === 'auto') {
-      return true
-    }
-    return isConnectionSecure(request)
+  function getShouldSaveSession (request, cookieId, saveUninitializedSession, rollingSessions) {
+    return cookieId !== request.session.encryptedSessionId
+      ? saveUninitializedSession || request.session.isModified()
+      : rollingSessions || (Boolean(request.session.cookie.expires) && request.session.isModified())
```
rclmenezes, line 201: I borrowed this code from `express-session`:
https://github.com/expressjs/session/blob/master/index.js#L464-L473

Interestingly, they have both `shouldSave` and `shouldSetCookie`. This distinction will be important later when we correctly implement `session.touch()`

rclmenezes:
And one more thing worth mentioning! Why did I abandon putting `cookie.expires.getTime()` in the session hash?

I observed that express-session doesn't do this:
https://github.com/expressjs/session/blob/master/index.js#L599-L602

After thinking about it, I realized that there's a difference between `isModified` and `isTouched`. The hash is purely used for checking modification.

So, if the session hasn't changed but we want to extend the TTL due to `rolling: true`, then we need to put this check in the plugin. Hence the changes here:
https://github.com/fastify/session/pull/149#discussion_r957829161

Hope this helps

lib/fastifySession.js
Uzlopak: I think the issue is, that even the session maybe expired we have now a session assigned to the request?!

rclmenezes: `destroy()` already replaces the session in the request:
https://github.com/fastify/session/blob/master/lib/session.js#L87-L90

I could avoid setting request.session later though, if it makes you feel more comfortable. But it shouldn't make a difference, I think

rclmenezes: I guess this could be a problem if `destroy()` runs into an error.

I pushed a commit that assigns the session later. Thanks!

lib/fastifySession.js
```diff
@@ -106,54 +110,49 @@ function session (fastify, options, next) {
...
-        decryptSession(sessionId, options, request, done)
+        decryptSession(cookieSessionId, options, request, done)
       }
     }
   }
 
   function onSend (options) {
     const cookieOpts = options.cookie
-    const cookieName = options.cookieName
-    const cookiePrefix = options.cookiePrefix
```
Uzlopak, line 125: Why is all the cookie prefix code gone?

lib/cookie.js
Uzlopak: I know that it is not the performance bottleneck, but how about we just store the timestamp instead of the date in the store?

lib/fastifySession.js
rclmenezes: I cleaned it up a bit. See this commit to see the isolated changes:
https://github.com/fastify/session/pull/149/commits/7eba911de77054fb0cf3503b7c18776cf50d51e1

After cleaning, I noticed that this line was a potential problem:
https://github.com/fastify/session/blob/master/lib/fastifySession.js#L143

`request.cookies[cookieName]` would still have the prefix so it shouldn't be compared to `encryptedSessionId` directly.

lib/cookie.js
rclmenezes: Unfortunately, `express-session` stores it as a Date object:
https://github.com/expressjs/session/blob/master/session/cookie.js#L53-L63

So this could be a potential issue if a service using `@fastify/session` and a service using `express-session` share a session store

lib/fastifySession.js
```diff
@@ -78,25 +84,23 @@ function session (fastify, options, next) {
...
-        session,
-        decryptedSessionId
-      )
+
+      request.session = restoredSession
       done()
     })
   }
 
+  function getCookieSessionId (request) {
```
Uzlopak, line 93: We could improve this:

```js
  const getCookieSessionId = hasCookiePrefix
    ? function getCookieSessionId(request) {
      const cookieValue = request.cookies[cookieName]
      return (
        cookieValue &&
        cookieValue.startsWith(cookiePrefix) &&
        cookieValue.slice(cookiePrefixLength)
      )
    }
    : function getCookieSessionId(request) {
      return request.cookies[cookieName]
    }

```

lib/fastifySession.js
```diff
@@ -192,19 +191,14 @@ function session (fastify, options, next) {
...
 
-  function shouldSaveSession (request, cookieOpts, saveUninitialized) {
-    if (!saveUninitialized && !request.session.isModified()) {
-      return false
-    }
-    if (cookieOpts.secure !== true || cookieOpts.secure === 'auto') {
-      return true
-    }
-    return isConnectionSecure(request)
+  function getShouldSaveSession (request, cookieId, saveUninitializedSession, rollingSessions) {
```
Uzlopak, line 198: getShouldSaveSession? why two verbs? shouldSaveSession seems still better.

the variable you instantiate in line 142 can be called saveSession or shouldSave

lib/fastifySession.js
Uzlopak: Should be always faster when we dont use a cookiePrefix

lib/fastifySession.js
```diff
@@ -106,54 +110,49 @@ function session (fastify, options, next) {
...
-      if (encryptedSessionId && hasCookiePrefix) {
-        encryptedSessionId = `${cookiePrefix}${encryptedSessionId}`
-      }
-
-      if (!shouldSaveSession(request, cookieOpts, saveUninitialized)) {
-      // if a session cookie is set, but has a different ID, clear it
-        if (request.cookies[cookieName] && request.cookies[cookieName] !== encryptedSessionId) {
+      const cookieSessionId = getCookieSessionId(request)
+      const shouldSaveSession = getShouldSaveSession(request, cookieSessionId, saveUninitializedSession, rollingSessions)
+      const isInsecureConnection = cookieOpts.secure === true && !isConnectionSecure(request)
```
Uzlopak, line 138: cookieOpts.secure can be also have the value 'auto'. I dont know if this is a good idea to just check for secure being true. Also secure could be set to undefined. So I have some doubts if this code is working as expected under any circumstances.

Uzlopak:
I think your PR fixes rolling. I have some remarks regarding the performance, but the cookie.opts secure remark worries me.

lib/fastifySession.js
rclmenezes: Hey man you're the one that said:

> Please dont shorten the variable name.

😛 

But point taken, will do

lib/fastifySession.js
rclmenezes: Yeah, I think this is a potential issue that already exists in the code. Probably for another PR. Looking at:

image

If you add a `!`:

```
!(cookieOpts.secure !== true || cookieOpts.secure === 'auto')
(cookieOpts.secure === true && cookieOpts.secure !== 'auto')
(cookieOpts.secure === true)
```

I think we should fix this in a separate ticket.

lib/fastifySession.js
Uzlopak: What about this?

lib/fastifySession.js
Uzlopak: We should create an issue after we merged this

Uzlopak (approved):

lib/fastifySession.js
rclmenezes: Oops sorry missed it

PR #160, opened 2022-09-06
rename session to fastifySession

Uzlopak:
Something which really annoyed me from the beginning :)

#### Checklist

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -5,7 +5,7 @@ const idGenerator = require('./idGenerator')()
 const Store = require('./store')
 const Session = require('./session')
 
-function session (fastify, options, next) {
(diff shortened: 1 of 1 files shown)


jsumners:
Please stop merging pull requests without review. In this particular case, I don't see a need for the change. We know it's a Fastify plugin. There's no need for the `fastify` prefix on the function name.

Uzlopak:
I think it is a useful change. We have another file called session.js where we have a Session-Class. Now I have a better clarity regarding naming inside.

zekth:
> I think it is a useful change. We have another file called session.js where we have a Session-Class. Now I have a better clarity regarding naming inside.

Even if you think that's a useful change, you should wait for a review.

Uzlopak:
ok

PR #161, opened 2022-09-06
Make it possible to call save() in decryptSession

### LINKED ISSUE #159: `session.save()` does not work on manually decrypted sessions
### Prerequisites

### Fastify version

4.5.3

### Plugin version

9.0.0

### Node.js version

16.14.0

### Operating system

Windows

11

### Description

I am trying to have a rolling session. ie, I create a session with a certain timestamp and then try to extend the session expiry inside a socket.io route. The web socket route retrieves the session by using `decryptSession` function of fastify and tries to call `session.touch()` and `session.save()` functions to extend the session at certain points inside the socket APIs and save them to my Redis store.

### Steps to Reproduce

Here's my code for this:

```
app.decryptSession(sid, req, () => {
      if (!req.session?.get('userInfo')?.username) {
        return rejectSocket();
      }
      req.session.touch();
      req.session.save();
      scheduleSessionTouch(socket, req.session);
      return resolve();
    });
```

Looks like the code errors out in the `save()` call with the error `Cannot read properties of undefined (reading 'set')` in the line `lib\session.js:139:39`

Is save() supposed to work on decrypted sessions?

### Expected Behavior

`save()` should store the session to Redis store without errors.

Uzlopak:
It seems that in some cases the request is not decorated with the sessionStore. Instead we keep now a reference to the sessionStore directly and access the sessionStore instead of always accessing it through the request-Object. 

#### Checklist

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -13,6 +13,7 @@ function fastifySession (fastify, options, next) {
 
   options = ensureDefaults(options)
 
+  const sessionStore = options.store
   const cookieSigner = options.signer
   const cookieName = options.cookieName
   const cookiePrefix = options.cookiePrefix
@@ -30,7 +31,7 @@ function fastifySession (fastify, options, next) {
     const cookie = { ...options.cookie, ...cookieOpts }
     decryptSession(sessionId, { ...options, cookie }, request, callback)
   })
-  fastify.decorateRequest('sessionStore', { getter: () => options.store })
+  fastify.decorateRequest('sessionStore', { getter: () => sessionStore })
   fastify.decorateRequest('session', null)
   fastify.addHook('onRequest', onRequest(options))
   fastify.addHook('onSend', onSend(options))
@@ -42,7 +43,13 @@ function fastifySession (fastify, options, next) {
 
     const unsignedCookie = cookieSigner.unsign(sessionId)
     if (unsignedCookie.valid === false) {
-      request.session = new Session(request, idGenerator, cookieOpts, cookieSigner)
+      request.session = new Session(
+        sessionStore,
+        request,
+        idGenerator,
+        cookieOpts,
+        cookieSigner
+      )
       done()
       return
     }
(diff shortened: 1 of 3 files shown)


SimenB:
I'm down with the `fastify.decorateRequest('sessionStore', { getter: () => sessionStore })` change, but I don't understand why it doesn't work. Can you add a test that fails without?

---

To me its seems the issue is that the `request` object is not the real one - it is accessed off of `socket.request`. Is that from `fastify-socket.io`? I have no idea, but to me the bug seems to be some other lib creating their own `Request` objects without copying decorated stuff properly.

Uzlopak:
Hmm. let me check

SimenB:
https://github.com/Shrihari-Prakash/fastify-session-bug/blob/58c67e2c2fb4fc3a4ed2cb576aefbbc416954465/index.js#L85 isn't a real `Request` - I don't think this is a bug in `@fastify/session`

Uzlopak:
Yeah, i agree with you. 

But maybe it makes anyway sense to pass and keep a reference to the sessionStore and use it. Making it this way I have the feeling that we separate it more and request does not need to be a fastifyRequest anymore. I am ok with it that the request is not a fastify request anymore. 

Will add now a unit test.

Uzlopak:
@SimenB 
It seems like we always made it possible to use a custom request object. 

https://github.com/fastify/session/blob/ec1276092f9f3fed76945f74e38b130c5486709a/test/session.test.js#L217

But we never tried if we can actually save the session when we use decryptSession.

So extended the unit test.

Uzlopak:
@SimenB 
Can you make a review please?

SimenB:
>should decryptSession with custom request object

Is this a use case we wanna support?

Uzlopak:
I went back to the tag of v6.0.0. And we...ll it seems like it never worked properly. 

So this is actually fixing it. 

I think there is no harm to make it possible.

Uzlopak:
@zekth 

It is not documented why decryptSession is deprecated. Only that you once requested to mark it as deprecated.

SimenB (approved):

Issue #163, opened 2022-09-06
Deprecating `decryptSession` breaks compatibility with web sockets.


shrihari-prakash:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

I see that in recent code changes, the functionality for manually decoding sessions using `decryptSession` is deprecated:
https://github.com/fastify/session/blob/ca0fbe6a7c5fa8ba9ad81616faf08e1bfc47a90e/lib/fastifySession.js#L24

However, this is an important functionality that could potentially render the package completely useless for hybrid services which use Fastify for REST APIs and Socket.IO for web socket end points. Typically, the requests coming to Socket.IO does not pass through Fastify's hooks which means there is no access to sessions inside web sockets at all.

For reference, even Fastify's own `@fastify/websocket` package does not use normal fastify `Request` object.

Is there a reason why this functions is deprecated? In my opinion, use cases where node services offer both REST and Socket APIs are extremely common and this use case should be supported. Or is there an alternative you would be introducing to share sessions between Fastify and other packages like Socket.IO if this is removed in future? This feature is extremely important for my integrations.

mcollina:
What would you recommend we do?

Shrihari-Prakash:
@mcollina is it possible to still retain this function in subsequent versions and not deprecate it if that's possible? Do you see why this should be deprecated in the first place? Does this function raise any security concerns?

mcollina:
Would you mind to send a PR and remove the deprecation notice?

Shrihari-Prakash:
Hi @mcollina , I've raised a PR to reverse the deprecation notice: #164 . Please provide your review.

PR #164, opened 2022-09-07
Remove deprecation notice for fastify-Instance decryptSession

Shrihari-Prakash:
Related to issue #163 .

The deprecation of `decryptSession` is reversed to support scenarios where node services can be hybrid and can potentially have the need to share the sessions between Fastify and other libraries like socket.io.

#### Checklist

- [X ] run `npm run test` and `npm run benchmark`
- [X ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -21,7 +21,6 @@ function fastifySession (fastify, options, next) {
   const cookiePrefixLength = hasCookiePrefix && cookiePrefix.length
 
   // Decorator function takes cookieOpts so we can customize on per-session basis.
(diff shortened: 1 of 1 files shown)


Uzlopak:
I realized that decryptSession is missing in the typescript typings. If this PR gets merged as we would agree to not deprecate the function, i would create a PR for adding the typings for decryptSession

mcollina (approved):
lgtm

PR #169, opened 2022-09-09
security: verify path of cookie/session

Uzlopak:
This PR ensures that we respect the path of the session. So we dont load the session if the cookie is sent to a different path. 

when is it relevant?

If I set the cookie to the path `/admin` and send it to `/admin-ui`, it was previously loading the session. 

#### Checklist

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -131,7 +131,7 @@ function fastifySession (fastify, options, next) {
       request.session = {}
 
       const url = request.raw.url
-      if (url.indexOf(cookieOpts.path || '/') !== 0) {
+      if (verifyPath(url, cookieOpts.path || '/') === false) {
         done()
         return
       }
@@ -228,6 +228,25 @@ function fastifySession (fastify, options, next) {
 
   function option (options, key, def) {
     return options[key] === undefined ? def : options[key]
+  }
+
+  function verifyPath (path, cookiePath) {
+    if (path === cookiePath) {
+      return true
+    }
+    const pathLength = path.length
+    const cookiePathLength = cookiePath.length
+
+    if (pathLength <= cookiePathLength) {
+      return false
+    } else if (path.startsWith(cookiePath)) {
+      if (path[cookiePathLength] === '/') {
+        return true
+      } else if (cookiePath[cookiePathLength - 1] === '/') {
+        return true
+      }
+    }
+    return false
   }
 }
 
--- a/test/verifyPath.test.js
+++ b/test/verifyPath.test.js
@@ -221,7 +221,7 @@ test('should handle path properly /5', async (t) => {
   t.equal(response2.statusCode, 200)
 })
 
-test('should handle path properly /5', async (t) => {
+test('should handle path properly /6', async (t) => {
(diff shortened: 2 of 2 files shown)


Uzlopak:
Maybe this is actually an issue of fastify-cookie?

test/verifyPath.test.js
```diff
@@ -0,0 +1,345 @@
...
+  fastify.get('/', (request, reply) => {
+    request.session.foo = 'bar'
+    reply.send(200)
+  })
+  fastify.get('/check', (request, reply) => {
+    t.equal(request.session.foo, 'bar')
+    reply.send(200)
+  })
+  await fastify.listen({ port: 0 })
+  t.teardown(() => { fastify.close() })
```
SimenB, line 28: why do we `listen` and `close` when we use `inject`?

lib/fastifySession.js
```diff
@@ -229,6 +229,25 @@ function fastifySession (fastify, options, next) {
...
+    if (path === cookiePath) {
+      return true
+    }
+    const pathLength = path.length
+    const cookiePathLength = cookiePath.length
+
+    if (pathLength <= cookiePathLength) {
+      return false
+    } else if (path.startsWith(cookiePath)) {
+      if (path[cookiePathLength] === '/') {
```
SimenB, line 243: do we need this `if`? seems to behave the same as the one 2 lines below

SimenB (approved):

lib/fastifySession.js
climba03003: > do we need this `if`? seems to behave the same as the one 2 lines below

I have through a while, but it actually does require.

The first assert `/docs` with `/docs/(.+)`, the second assert `/docs/` with `/docs/(.+)`.

Uzlopak: E.g. 

```
path = /check/
cookiePath = /check
```

`path[cookiePathLength]` would be '/' 

So we know that the path is terminated with a '/'

The below one covers another case

```
path = /check/index
cookiePath = /check/
```

So we know from line 242 that path startsWith cookiePath. So we know that path starts with `/check/`. But the line 243 would return false, because `path[cookiePathLength]` would be `i` ( /check/**i**ndex). So with line 245 we check if cookiePath is terminated with a '/' and if so we can say that the path is valid.

test/verifyPath.test.js
Uzlopak: I never thought about it. I just copy pasted other tests and thought I have to use listen to be able to do injects. If it is not the case, than maybe we can clean this up in follow up PR.

lib/fastifySession.js
SimenB: cool, thanks for thorough explanation 👍

test/verifyPath.test.js
Uzlopak: Modified the new unit tests. But we should make a follow up PR removing it from the other places too, if applicable.

Uzlopak:
Ok to be merged?

mcollina (approved):
lgtm

PR #170, opened 2022-09-09
Dont load sessions from insecure connections from secure cookies and vice versa

Uzlopak:
I dont know how to explain this properly. I think this is not a real security issue, but should be anyway handled correctly:

If you send a cookie with secure flag, through an insecure http connection, it would  result in loading the session. I would not expect that. 

We are signing and unsigning cookies. So the content of a cookie is practically safe. So that is the reason why it is not a real security issue. 

But it is possible to send a secure cookie to an unsecure connection. So it loads the session and in the handler we can access the session, even though the connection was not secure anyway. So in the handler you can modify the session and call .save or because you loaded the data you process it anyway.

Maybe it is a security breach because you are a bank or whatever and need even to handle that case very strict because of some security rules/laws. 

Maybe it is a brainfart from my side and I make it more complicated and didnt understand how cookies should be treaded.

#### Checklist

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -29,7 +29,13 @@ function fastifySession (fastify, options, next) {
     }
 
     const cookie = { ...options.cookie, ...cookieOpts }
-    decryptSession(sessionId, { ...options, cookie }, request, true, callback)
+    decryptSession({
+      sessionId,
+      options: { ...options, cookie },
+      request,
+      custom: true,
+      callback
+    })
   })
   fastify.decorateRequest('sessionStore', { getter: () => sessionStore })
   fastify.decorateRequest('session', null)
@@ -37,7 +43,14 @@ function fastifySession (fastify, options, next) {
   fastify.addHook('onSend', onSend(options))
   next()
 
-  function decryptSession (sessionId, options, request, custom, done) {
+  function decryptSession (opts) {
+    const {
+      sessionId,
+      options,
+      request,
+      custom,
+      callback: done
+    } = opts
     const cookieOpts = options.cookie
     const idGenerator = options.idGenerator
 
@@ -152,7 +165,13 @@ function fastifySession (fastify, options, next) {
         )
         done()
       } else {
-        decryptSession(cookieSessionId, options, request, false, done)
+        decryptSession({
+          sessionId: cookieSessionId,
+          options,
+          request,
+          custom: false,
+          callback: done
+        })
       }
     }
   }
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -29,7 +29,7 @@ function fastifySession (fastify, options, next) {
     }
 
     const cookie = { ...options.cookie, ...cookieOpts }
-    decryptSession(sessionId, { ...options, cookie }, request, callback)
+    decryptSession(sessionId, { ...options, cookie }, request, true, callback)
   })
   fastify.decorateRequest('sessionStore', { getter: () => sessionStore })
   fastify.decorateRequest('session', null)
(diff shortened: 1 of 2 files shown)


lib/fastifySession.js
```diff
@@ -29,15 +29,15 @@ function fastifySession (fastify, options, next) {
...
+    decryptSession(sessionId, { ...options, cookie }, request, true, callback)
   })
   fastify.decorateRequest('sessionStore', { getter: () => sessionStore })
   fastify.decorateRequest('session', null)
   fastify.addHook('onRequest', onRequest(options))
   fastify.addHook('onSend', onSend(options))
   next()
 
-  function decryptSession (sessionId, options, request, done) {
+  function decryptSession (sessionId, options, request, custom, done) {
```
SimenB, line 40: note that this is a breaking change, and I'd expect at least type definitions to be wrong? Should also be called out in the changelog somehow, and not just the title of the PR. v10 will be full of breaking changes, and I think migrating from v9 will get hard without some sort of upgrade guide

lib/fastifySession.js
```diff
@@ -105,7 +105,12 @@ function fastifySession (fastify, options, next) {
         return
       }
 
-      request.session = restoredSession
+      const secureConnectionCookieMismatch = custom !== true &&
+        (typeof cookieOpts.secure === 'boolean'
+          ? cookieOpts.secure !== isConnectionSecure(request)
+          : session.cookie.secure !== isConnectionSecure(request))
+
+      request.session = secureConnectionCookieMismatch ? {} : restoredSession
```
SimenB, line 113: should it be an empty object and not a new `Session` instance?

lib/fastifySession.js
jsumners: Ugh, what a terrible signature. Make it an object.

lib/fastifySession.js
Uzlopak: I dont know. I am open for discussion.

lib/fastifySession.js
Uzlopak: This decryptSession is not external API. So should be not breaking.

SimenB: >This decryptSession is not external API. So should be not breaking.

Ah, fair 👍 

---

I still think we need an upgrade guide, but I agree _this_ is internal API

mcollina (changes requested):
I'm -1 for this. Most of the time HTTPS is terminated at the gateway/reverse proxy level, so the current connection is plain text but running within a private/encrypted network.
This change will make using this plugin almost impossible for the most complex deployments, as well as making local development exceptionally hard.

I think we should:

1. have an option to configure this behavior
2. by default this should log at _warn_ level
3. it could be configured to ignore the problem or avoid loading the session

lib/fastifySession.js
SimenB: If people do e.g. `request.session.save` this will fail now and https://github.com/fastify/session/blob/ed4f9ac14cca8c94f866aaa70fa63e5235fbaf33/types/types.d.ts#L14 is wrong

Uzlopak:
@mcollina 

I dont know if ingress/nginx/traefik filters out cookies based on the Secure flag. But I would assume that they set the `x-forwarded-proto` to `https`.

I was just curios, about the fact regarding this line:

https://github.com/fastify/session/blob/ed4f9ac14cca8c94f866aaa70fa63e5235fbaf33/lib/fastifySession.js#L169

So we are NOT automatically saving the session if we have configured a secure cookie but the connection is not secure. So my thought was: Why do we make the security at the very last moment? We could have the effect that somebody makes all the changes in the session because he thinks that the session would be automatically saved. And without any notice, it will not be stored. 

And then I assume a dev would think: "Maybe it does not automatically save the session." So what they do is to call `request.session.save()` and suddenly a cookie, which should have never been generated nor should have been saved is tainting the database and the behaviour of the code. 

But I am ok not to merge this. I close this for now and will sleep over it.

mcollina:
You might be correct, this is a complex module and this might not be a problem. However let's make sure this won't impact some
basic cases first.

Issue #173, opened 2022-10-08
Session plugin sets an empty SID cookie for a wrong domain


TheWashiba:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.7.0

### Plugin version

10.0.0

### Node.js version

18.x

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

Ubuntu 20.04

### Description

I have an interesting case of an extra SID cookie being set for a subdomain.
My Fastify API app is server behind a proxy under a subdomain `api.development.local`, while the session cookie domain is set as `development.local`. 

When I hit a request that saves some state:
```typescript
request.session.user = { id, username, permissions, state: 'authenticated' };
await request.session.save();
```
Any subsequent request will successfully retrieve it by
```typescript
request.session.user
``` 
But after hitting an API with
```typescript
await request.session.destroy();
```
The HTTP request does not clear the SID cookie. Shouldn't it be cleared after it? When I access a session in a subsequent request, it adds an empty cookie, but there's is a problem - different domains:
1. Old SID Cookie with domain `development.local`
2. New empty SID cookie with domain `api.development.local`

Subdomain specific empty cookie now takes precedence and will always be sent to the API server, rendering the session plugin unable to work properly. I believe the issue lies in declared domain not used on clearing the cookie, so the default incoming requests domain is used (with subdomain included).

### Steps to Reproduce

In description.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

PR #174, opened 2022-10-10
fix: set domain when clearing cookie

TheWashiba:
This PR fixes the issue #173. If user specifies `domain` in a cookie, but runs the server behind a subdomain, then a duplicate empty cookie is created at the specified condition and it will always be used for the following HTTP requests as a SID cookie, breaking the session mechanism.

#### Checklist

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -170,7 +170,7 @@ function fastifySession (fastify, options, next) {
       if (!saveSession || isInsecureConnection) {
         // if a session cookie is set, but has a different ID, clear it
         if (cookieSessionId && cookieSessionId !== session.encryptedSessionId) {
-          reply.clearCookie(cookieName)
+          reply.clearCookie(cookieName, { domain: cookieOpts.domain })
         }
         done()
         return
--- a/test/session.test.js
+++ b/test/session.test.js
@@ -723,6 +723,22 @@ test("clears cookie if not backed by a session, and there's nothing to save", as
   t.plan(2)
   const fastify = await buildFastify((request, reply) => {
     reply.send(200)
+  }, DEFAULT_OPTIONS)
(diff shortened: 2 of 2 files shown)


TheWashiba:
> Why did you patch the test and not wrote a new one?

Seemed like an appropriate way as it was about correcting the clear cookie behavior, but yeah, I can create a separate test case for the domain option only.

mcollina (approved):
lgtm

Eomm (approved):

Uzlopak:
LOL. I was on the way to the office to review it.

Eomm:
I read your comment about the test being solved.

Please add any comments, and I will take care of them

Uzlopak:
I had only two thoughts:

1. could we do at about line 159 

```js
const clearCookieOpts = cookieOpts.domain
  ? { domain: cookieOpts.domain }
  : { }

```
and use it instead as clearCookie option like

```js
reply.clearCookie(cookieName, clearCookieOpts)
```

2. Do we have some more underlying issues like missing path? Should we not just pass cookieOpts directly into the clearCookie?
```js
reply.clearCookie(cookieName, cookieOpts)
```

I think it needs some investigations.

Eomm:
> could we do at about line 159

It does not change the output: `{ domain: undefined }` is evicted.

> Do we have some more underlying issues like missing path? 

yes, it could be the case 👍🏽

Uzlopak:
I just wanted to avoid generating the object over and over again.

Eomm:
Yeah, the old dilemma of putting objects in the Old Space or New Space
https://dev.to/eomm/garbage-collector-stress-242f

The question is: how many times do we run that code?

Uzlopak:
idk but i would prematurely optimize it anyway.

PR #177, opened 2022-10-19
fix: align cookie implementation with express-session

### LINKED ISSUE #176: What is the reason behind converting the `cookie.maxAge` prop to an `Expires` value in the actual cookie?
### Prerequisites

- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

`getCookieOpts()` always converts the `maxAge` property to the `expires` property by creating a new Date and adding the `maxAge` to the current server's time. The returned object omits the `maxAge` property then. I believe this creates issues if the client's time is out of sync and goes against the expected behavior.

What is the rationale behind this design choice? To be honest, I hope the answer is not just "because express does it the same way" 😅

climba03003:
Fixes #176 

#### Checklist


Uzlopak:
I dont think this is correct. originalMaxAge is used internally.

@rclmenezes

climba03003:
> I dont think this is correct. originalMaxAge is used internally.

It still used internally, nothing is changed.

SimenB (approved):
I think this makes sense 👍

rclmenezes:
LGTM

@Uzlopak the key comment is here: https://github.com/fastify/session/issues/176#issuecomment-1283921462

I confirmed myself that express-session does _not_ set `expired` every time in the store, so this behavior is consistent.

Also, looks like express-session uses originalMaxAge the same way:
https://github.com/expressjs/session/search?q=originalmaxage

Nice PR

Uzlopak (approved):

PR #180, opened 2022-12-20
Allow null Session in SessionCallback when there is an error.

kylerush:
It looks like the existing code support an error being thrown  on `store.get`, however, the type definition `SessionCallback` for the callback function on `store.get` requires a `Session` type variable for the session argument. This change would not cause the type check to fail when you call the `store.get` callback function like so:

```
callback(new Error("Some error"), null)
```

#### Checklist

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -26,7 +26,12 @@ type FastifySession = FastifyPluginCallback<fastifySession.FastifySessionOptions
 }
 
 type Callback = (err?: Error) => void;
-type CallbackSession = (err: Error | null, result: Fastify.Session) => void;
+//type CallbackSession = (err: Error | null, result: Fastify.Session) => void;
+
+interface CallbackSession {
+  (err: null, result: Fastify.Session): void
+  (err: Error, result: null): void
+}
 
 interface SessionData extends ExpressSessionData {
   sessionId: string;
--- a/types/types.test-d.ts
+++ b/types/types.test-d.ts
@@ -88,8 +88,10 @@ app.route({
     request.sessionStore.set('session-set-test', request.session, () => {});
     request.sessionStore.get('', (err, session) => {
       expectType<Error | null>(err);
-      expectType<Session>(session);
-      expectType<{ id: number } | undefined>(session.user);
+      expectType<Session | null>(session);
+      if (session) {
+        expectType<{ id: number } | undefined>(session.user);
+      }
     });
     expectType<void>(request.session.set('foo', 'bar'));
     expectType<string>(request.session.get('foo'));


types/types.d.ts
```diff
@@ -26,7 +26,12 @@ type FastifySession = FastifyPluginCallback<fastifySession.FastifySessionOptions
 }
 
 type Callback = (err?: Error) => void;
-type CallbackSession = (err: Error | null, result: Fastify.Session) => void;
+//type CallbackSession = (err: Error | null, result: Fastify.Session) => void;
+
+interface CallbackSession {
+  (err: null, result: Fastify.Session): void
+  (err: Error, result: null): void
```
climba03003, line 33: Just don't pass anything when there is error.
```suggestion
  (err: Error): void
```

climba03003 (changes requested):

kylerush:
Thanks @climba03003! I'm going to commit your suggestion and then would like to propose one more change to export the `SessionCallback` type. Will add that change shortly.

types/types.d.ts
```diff
@@ -26,7 +26,12 @@ type FastifySession = FastifyPluginCallback<fastifySession.FastifySessionOptions
 }
 
 type Callback = (err?: Error) => void;
-type CallbackSession = (err: Error | null, result: Fastify.Session) => void;
+//type CallbackSession = (err: Error | null, result: Fastify.Session) => void;
+
+interface CallbackSession {
```
Uzlopak, line 31: I would have expected a type and not an interface:

type CallbackSession = ((err: null, result: Fastify.Session) => void) | ((err: Error) => void);

kylerush: Oh, this is cool. I didn't know you could do types like this. I switched it to a type, though, with a different syntax. Let me know if you would like me to change it to this type syntax.

mcollina (approved):
lgtm

Uzlopak:
@climba03003 PTAL

types/types.d.ts
```diff
@@ -80,6 +79,8 @@ interface Signer {
 }
 
 declare namespace fastifySession {
+  export type CallbackSession = (err: null | Error, result: undefined | Fastify.Session) => void
```
Uzlopak, line 82: Wouldnt this be better?
```suggestion
  export type CallbackSession = ((err: Error, result: undefined) => void) | ((err: null, result: undefined | Fastify.Session) => void)
```

climba03003:
```suggestion
  export type CallbackSession = ((err: Error) => void) | ((err: null, result: Fastify.Session) => void)
```

kylerush: @Uzlopak - I think what @climba03003 suggested is best. We don't need to return an undefined result, we just don't return anything at all, which makes `result` undefined.

kylerush: @Uzlopak - I think what @climba03003 suggested is best. We don't need to return an undefined result, we just don't return anything at all, which makes `result` undefined.

climba03003 (approved):

kylerush:
@climba03003 this new type definition seems to not be working, or at least I don't understand how it should work.

With the suggestion change, typescript tells me:

1. `Argument of type 'null' is not assignable to parameter of type 'Error'.`
2. `Argument of type 'Session' is not assignable to parameter of type 'undefined'.`

Screenshot 2022-12-26 at 11 51 43 AM

I am not quite sure how to solve this problem. Is seems this union type definition doesn't quite work for the use case. Any idea?

I tried to change it to an overloaded function, but that caused the arguments the callback function accepts to lose their types.

mcollina:
CI is not happy

kylerush:
Yeah it needs a little more work. The new tuple type seems to be causing issues as currently implemented. Will try to knock this out tomorrow.

sarneeh:
Hey folks 👋 I see this one didn't get finished and I'd like to have it fixed as well.

I took some time to tinker around with `CallbackSession` type and found out that we don't really have a flawless solution for this case. Here are a few implementations that I tested out:

### [Intersection](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwghgGwQIzgYwNZQLxQBT54QBOxAXFAKKkD2xAlDgHxQBuNAlgCaMBkhJclAB2AVyQAaKMQgBnccApiAtshKNsLdt0b0A3ACgDaGsNnAoAcwjAAynNkdTOfGkQp0GCvCSpMGlgBvAyhQqDdfTzxhCAB3KlpiPHp9KAB6NPC6GTRgELCIj0xouITiOmSpAFZUjKhgAAty2NkoQTopE1IIXPzQwr8MaPEEatrMrpy8gF8jazsHJ2E8IkSAqGCw9MywOGI4ZRsSVrByyGIEEHrwCC4pABVbLOUwDgQ5PukbUWJhQ2n6HMbPZZI5TCt2sQpDJ5AhgOtNmE6rt9odgMcoKcaOdLtdIHcoMhRBZHlAuDQ5CIaBYTC83hBPjJgD8-gYAUA)
```ts
type Callback = ( ((err: Error) => void) & ((err: null, result: number) => void) );

const getSession = (callback: Callback) => {
    callback(new Error()); // correct
    callback(new Error(), 5); // throws error, correct
    callback(null, 5); // correct
}

getSession((error) => {
    // parameters properly typed, TS compiles
    return;
})

getSession((error, result) => {
    // parameters properly typed, but TS does not compile
    return;
})
```

### [Function overload](https://www.typescriptlang.org/play?#code/C4TwDgpgBAwghgGwQIzgYwNZQLxQN4BQUxUAFBAE4UBcUAolQPYUCUtAbowJYAmA3ERLkqtAHYBXJABooFCAGdJwMeIC2ySmyideAgL4ECaRqPnAoAcwjAAygvlcTOMmkQp0GWvCSpMLHAB8+ILErj4epKIQAO70TBSkLCx8UAD0qVDGVBBowCGZbr4YkTFxFMyJMgCsyWkZwAAW5dHyUJTlFDJZcrn5Ye6YkZII1bXpmcw9eQYEVrb2jqKkwh3+2EGEJHVQYHAUcKrWlK1g5ZAUCCBQoJA8MgAqNhOqYFwICvlywOIUovoshjmdnkDhMy3azBkckUCGAaw2+XGu32h2Axx2Z0ol2u4AgdygyHE5keUB4jAUUFEjHMxhebwgn2sPz+BD0LCAA)
```ts
type Callback = {
    (err: Error): void;
    (err: null, result: number): void;
}

const getSession = (callback: Callback) => {
    callback(new Error()); // correct
    callback(new Error(), 5); // throws error, correct
    callback(null, 5); // correct
}

getSession((error) => {
    // parameters properly typed, TS compiles
    return;
})

getSession((error, result) => {
    // parameters properly typed, but TS does not compile
    return;
})
```

You can see within the comments that those do not work as intended due to how TypeScript works. I think that the most optimal way we could go is this:

```ts
type Callback = {
    (err: Error, result?: undefined): void;
    (err: null, result: number): void;
}

const getSession = (callback: Callback) => {
    callback(new Error()); // correct
    callback(new Error(), 5); // throws error, correct
    callback(null, 5); // correct
}

getSession((error) => {
    // parameters properly typed but are not fully precise, TS compiles
    return;
})

getSession((error, result) => {
    // parameters properly typed but are not fully precise, TS compiles
    return;
})
```

This way we lose the preciseness of the arguments of the callback that we pass to the `get` function, but at least `callback` calls work as intended.

@kylerush Let me know if you have time to finish this off. If not, please let me know - I'll create a separate PR with the fix.

Let me know what you folks think!

PR #182, opened 2022-12-22
types: fix Session interface reference

### LINKED ISSUE #181: request.session type is always any
### Prerequisites

### Fastify version

4.10.2

### Plugin version

10.1.0

### Node.js version

18.12.1

### Operating system

macOS

13.1

### Description

Even after adding a type declaration like
```typescript
declare module "fastify" {
    interface Session {
        user: User
    }
}
```
to the code the type of request.session is still `any`.

Screenshot 2022-12-22 at 10 56 22 AM

### Steps to Reproduce

start an empty fastify project, add fastify-session and follow https://github.com/fastify/session#typescript-support

### Expected Behavior

the session type should be applied to request.session

FabianFrank:
now Session actually resolves to the session interface defined in the same files instead of being an unresolved any. goal of this change is to fix #181 

before:
Screenshot 2022-12-22 at 11 53 10 AM

after:
Screenshot 2022-12-22 at 11 53 22 AM

#### Checklist

This doesn't require a new test/benchmark or doc change to the best of my knowledge.

### DIFF
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -11,7 +11,7 @@ declare module 'fastify' {
 
   interface FastifyRequest {
     /** Allows to access or modify the session data. */
-    session: Fastify.Session;
+    session: Session;
 
     /** A session store. */
     sessionStore: Readonly<fastifySession.SessionStore>;


Uzlopak:
Actually we could need a typing test

FabianFrank:
@Uzlopak there's already a typing test for this at https://github.com/fastify/session/blob/6462ff40acfad405f6898d9b4ee06890902d903a/types/types.test-d.ts#L92 and the test doesn't break with this change. it's unfortunate that the types work under test but not when installed as a dependency in another project. frankly, I'm not familiar enough with tsd to figure out how to make the test fail with `Fastify.Session`.

Uzlopak:
I will have a look later

Eomm (approved):

mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test? We use tsd for typescript.

FabianFrank:
@mcollina please see https://github.com/fastify/session/pull/182#issuecomment-1363317191

mcollina:
I'm uncomfortable in landing this without a test.

FabianFrank:
@mcollina how do you think the existing test for this type should change?

mcollina:
I don't know. I'm not the one proposing the change in this PR.

climba03003 (approved):
LGTM.

Let me explain a bit of the problem. 
The local assertion will always pass since as a package point, we know `Fastify` types before using it.
However, when it comes to the consumer. It will create an infinite loop of referencing. So, it will return `any` as a result.

For a test, we need to create a dummy package.json, install itself as dependent. Then, assertion the types inside the dummy package. Would it be too heavy as a test?

FabianFrank:
In https://github.com/fastify/session/issues/181#issuecomment-1365394199 we were able to track this down to the `esModuleInterop` setting in typescript, so to build on @climba03003's idea of a separate test package, we'd also have to make that test package an ESM module and disable module interop. Then the question is are we comfortable to only test the ESM case or do we want to test the CJS case as well, i.e. have 2 test packages?

mcollina (approved):
lgtm

PR #189, opened 2023-03-20
feat: add ignoreFields option to regenerate method

RafaelGSS:
You can pass an array of fields that should be kept when the session is regenerated

### DIFF
--- a/README.md
+++ b/README.md
@@ -118,7 +118,7 @@ Allows to destroy the session in the store. If you do not pass a callback, a Pro
 
 Updates the `expires` property of the session's cookie.
 
-#### Session#regenerate(callback)
+#### Session#regenerate([ignoreFields, ]callback)
 
 Regenerates the session by generating a new `sessionId` and persist it to the store. If you do not pass a callback, a Promise will be returned.
 ```js
@@ -132,6 +132,8 @@ fastify.get('/regenerate', (request, reply, done) => {
   });
 });
 ```
+
+You can pass an array of fields that should be kept when the session is regenerated
 
 #### Session#reload(callback)
 
--- a/lib/session.js
+++ b/lib/session.js
@@ -60,16 +60,26 @@ module.exports = class Session {
     }
   }
 
-  regenerate (callback) {
-    if (callback) {
-      const session = new Session(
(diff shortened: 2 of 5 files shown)


types/types.d.ts
```diff
@@ -38,9 +38,13 @@ interface SessionData extends ExpressSessionData {
 
   /**
    * Regenerates the session by generating a new `sessionId`.
+   *
+   * ignoreFields specifies which fields should be kept in the new session object.
    */
   regenerate(callback: Callback): void;
+  regenerate(callback: Callback, ignoreFields: string[]): void;
```
Eomm, line 45: Isn't the callback as the first parameter odd?

README.md
```diff
@@ -118,7 +118,7 @@ Allows to destroy the session in the store. If you do not pass a callback, a Pro
 
 Updates the `expires` property of the session's cookie.
 
-#### Session#regenerate(callback)
+#### Session#regenerate(callback, [ignoreFields])
```
climba03003, line 121: `callback` should place at the last field.
We can check if the first field is `Array` or `Function` to prevent breaking changes.

climba03003 (changes requested):

types/types.d.ts
RafaelGSS: yes, but it will not introduce a breaking chance

climba03003 (approved):

lib/session.js
```diff
@@ -85,6 +96,12 @@ module.exports = class Session {
           this[cookieSignerKey]
         )
 
+        if (Array.isArray(keys)) {
+          for (const key of keys) {
+            session.set(key, this[key])
+          }
+        }
```
mcollina, line 103: I don't really like this code duplication. I think we should move the whole `new Session` block outside of the if-else.

mcollina (changes requested):

mcollina (approved):
lgtm but CI is failing

RafaelGSS:
> lgtm but CI is failing

I've run it several times locally and it passes. I've re-run it on actions and it's passing as well. It doesn't seem like a test flaky, but an environment flaky (gh actions).

PR #190, opened 2023-03-22
fix: express session type compatibility

ojeytonwilliams:
#### Checklist

This PR was motivated by the current need to assert that the Store types are `any` due to the incompatibility of @fastify/session's types with express-session's. e.g.

```ts
server.register(fastifySession, {
  store: new RedisStore({
    client: redisClient,
  }) as any,
```

I tried to keep the changes to a minimum, but it was necessary to alter the callbacks types and to change precisely what was in the various Session types.

The callback needed updating because express-session's callbacks all define their errors as `any` and because we can't guarantee that a `session` is passed to the callback.

The Session types needed updating because the stores are only compatible with express-session's `SessionData` rather than the full set of properties that a fastify session object has.

I wasn't able to run the benchmarks locally, this PR only modifies types, so I hope that's okay.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,27 +178,6 @@ declare module "fastify" {
 }
 ```
 
-While this plugin can be used with express-session compatible stores, the type definitions of some stores might be tied to express-session, which means that casting to `any` might be required. For example:
-
-```ts
-import fastifySession from '@fastify/session'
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

PR #192, opened 2023-03-31
fix: separate an actual path from url search part before verification

ohchi:
I encountered a problem that cookie.path option didn't work if:
1. Path was longer then just "/".
2. Get request had a query parameters.


mcollina (approved):
lgtm

PR #196, opened 2023-04-11
feat!: stricter types

### LINKED ISSUE #194: Allow to override FastifyRequest['session']
### Prerequisites

- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Allow to override FastifySessionObject or parts of it, e.g. I would like to be able to do this:

```ts
type SessionData = {
  sessionId?: string;
  tikTokOAuthState?: string;
  userAccountAuthenticationTokenId?: AuthenticationToken;
  userAccountId?: number;
};

declare module 'fastify' {
  interface FastifyRequest {
    session: FastifySession<SessionData>;
  }

  interface Session {
    get: <T extends string>(key: T) => SessionData[T];
    set: <T extends string>(key: T, value: SessionData[T]) => void;
  }
}
```

As far as I can see, I am not even able to override `FastifyRequest` `session`.

### Motivation

Have typed `get`, `set`, etc. methods.

### Example

_No response_

climba03003:
Fixes #194 

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,6 +180,16 @@ declare module "fastify" {
 }
 ```
 
+When you think that the getter or setter is too strict.
+You are allowed to use `any` types to loosen the check.
+
+```ts
+fastify.get('/', async function(request) {
+  request.session.get<any>('not-exist')
+  request.session.set<any>('not-exist', 'happy')
+})
+```
+
 ## License
 
 [MIT](./LICENSE)
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -87,14 +87,15 @@ declare namespace fastifySession {
     save(): Promise<void>;
 
     /** sets values in the session. */
-    set(key: string, value: unknown): void;
+    set<K extends keyof Fastify.Session, V = Fastify.Session[K]>(key: K, value: V): void;
 
     /** gets values from the session. */
-    get<T>(key: string): T;
+    get<K extends keyof Fastify.Session, V = Fastify.Session[K] | undefined>(key: K): V;
 
     /** checks if session has been modified since it was generated or loaded from the store. */
     isModified(): boolean;
   }
+
   export interface SessionStore {
     set(
       sessionId: string,
--- a/types/types.d.ts
+++ b/types/types.d.ts
@@ -1,7 +1,7 @@
 /// <reference types='node' />
 
-import { FastifyPluginCallback } from 'fastify';
 import type * as Fastify from 'fastify';
+import { FastifyPluginCallback } from 'fastify';
 
 declare module 'fastify' {
   interface FastifyInstance {
@@ -11,7 +11,7 @@ declare module 'fastify' {
 
   interface FastifyRequest {
     /** Allows to access or modify the session data. */
-    session: FastifySessionObject;
+    session: fastifySession.FastifySessionObject;
 
     /** A session store. */
     sessionStore: Readonly<fastifySession.SessionStore>;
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

climba03003:
This PR can be split into two part.
The first commit is already enough for override the types.
The second commit is stricter typing which may break some of the user if they are doing `request.session.get<'my-value'>('something')`

gajus:
What's the reason `FastifySessionObject` needs to be in `types/types.d.ts` or in a `fastifySession` namespace to begin with?

Why mess with globals instead of just exporting `FastifySessionObject` type?

If we are making a breaking change, I'd rather see something that allows me to do:

```ts
import { type FastifySession } from '@fastify/session';

declare module 'fastify' {
  interface FastifyRequest {
    session: FastifySession<SessionData>;
  }
}
```

rather than this being done in `.d.ts` file that comes with the dependency.

This would also make it easier to patch type issues in the future.

gajus:
> I have no idea what you means by mess with globals.

```ts
declare module 'fastify' {
  interface FastifyInstance {
    decryptSession<Request extends Record<string, any> = FastifyRequest>(sessionId: string, request: Request, cookieOpts: fastifySession.CookieOptions, callback: Callback): void;
    decryptSession<Request extends Record<string, any> = FastifyRequest>(sessionId: string, request: Request, callback: Callback): void;
  }

  interface FastifyRequest {
    /** Allows to access or modify the session data. */
    session: FastifySessionObject;

    /** A session store. */
    sessionStore: Readonly<fastifySession.SessionStore>;
  }
  
  interface Session extends ExpressSessionData { }
}
```

All of this code is effectively overriding global fastify types.

I suggest to remove all of this code entirely and allow users to define this in their project.

> It moves inside the namespace so you are allowed to override the object.
> Without placing in the namespace, you will never been able to do any action to it.

Move it out of the namespace and just export `FastifySessionObject` type.

This way I could (in my own project) utilize it such as shown earlier:

```ts
import { type FastifySession } from '@fastify/session';

declare module 'fastify' {
  interface FastifyRequest {
    session: FastifySession<SessionData>;
  }
}
```

climba03003:
> Simply installing a dependency should not auto-magically change types of fastify.

Installing `dependency` do not change the types.
Requiring the `dependency` does merge the types. 
Most people are expecting out-of-box experience more than manual configuration.

> Move it out of the namespace

I have no idea how you conclude that it should move it out of `namespace`.
We are using `export =` signature means there will be only `namespace` exporting.

Your example code is also **REQUIRED** to place inside `namespace` in order to works.

gajus:
> Most people are expecting out-of-box experience more than manual configuration.

Feels like generalization promoting bad practices.

I really do not expect/want that depending on a dependency changes my fastify types.

Just a few extra lines of code to avoid 'magic' behavior that is hard to patch seems like a good trade.

climba03003:
Declaration merging is explained inside the TypeScript document.
It is a valid usage of TypeScript and not something so called "magic".

Actually, since I am the contributor of fastify. You are the only one complaining on using declaration merging by default and claiming it as bad practice.

mcollina (approved):
lgtm

mcollina:
@climba03003 can this land?

climba03003:
> can this land?

Personally, yes.

But I would like to here voice from @fastify/typescript and see if they oppose.

fox1t (approved):
LGTM.
This is definitely a good improvement.

climba03003:
> @climba03003 should this ship as minor?

Yes

PR #201, opened 2023-07-03
fix: manually persisting the session should make `isModified` return false

SimenB:
#### Checklist

Following up #80. We're seeing some data loss after updating the session in a distributed system, and the suspect is `onSend` in this library storing old data, overwriting the update performed by either a different concurrent request, or a request handled by another pod.

Note that this is slightly different from what `express-session` does.

https://github.com/expressjs/session/blob/1010fadc2f071ddf2add94235d72224cf65159c6/index.js#L403-L407

It tracks `modified` and `saved` separately. I'm not sure if that is necessary?

### DIFF
--- a/test/session.test.js
+++ b/test/session.test.js
@@ -898,11 +898,58 @@ test('when rolling is false, only save session when it changes', async t => {
   const response2 = await fastify.inject({ path: '/', headers: { cookie: `sessionId=${sessionId}` } })
   const setCookieHeader2 = response2.headers['set-cookie']
 
-  t.equal(response1.statusCode, 200)
+  t.equal(response2.statusCode, 200)
   // still only called once
   t.equal(setCount, 1)
   // no set-cookie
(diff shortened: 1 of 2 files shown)


test/session.test.js
```diff
@@ -898,13 +898,60 @@ test('when rolling is false, only save session when it changes', async t => {
   const response2 = await fastify.inject({ path: '/', headers: { cookie: `sessionId=${sessionId}` } })
   const setCookieHeader2 = response2.headers['set-cookie']
 
-  t.equal(response1.statusCode, 200)
+  t.equal(response2.statusCode, 200)
```
SimenB, line 901: noticed my old test checked the wrong object 😅

mcollina (approved):
lgtm

SimenB:
This _might_ have caused a regression where the cookie isn't set as it should in the `onSend`. We have some unit tests failing at work.

I won't be able to look into it until tomorrow

PR #227, opened 2023-12-06
Fix: maxAge description in readme

likerRr:
There is a mistake in `maxAge` description - should be `seconds` instead of `milliseconds`

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Prefix for the value of the cookie. This is useful for compatibility with `expre
 ##### cookie
 The options object is used to generate the `Set-Cookie` header of the session cookie. May have the following properties:
 * `path` - The `Path` attribute. Defaults to `/` (the root path).
-* `maxAge` - A `number` in milliseconds that specifies the `Expires` attribute by adding the specified milliseconds to the current date. If both `expires` and `maxAge` are set, then `maxAge` is used.
+* `maxAge` - A `number` in seconds that specifies the `Expires` attribute by adding the specified seconds to the current date. If both `expires` and `maxAge` are set, then `maxAge` is used.
 * `httpOnly` - The `boolean` value of the `HttpOnly` attribute. Defaults to true.
 * `secure` - The `boolean` value of the `Secure` attribute. Set this option to false when communicating over an unencrypted (HTTP) connection. Value can be set to `auto`; in this case, the `Secure` attribute will be set to false for an HTTP request. In the case of HTTPS, it will be set to true.  Defaults to true.
 * `expires` - The expiration `date` used for the `Expires` attribute. If both `expires` and `maxAge` are set, then `maxAge` is used.


gurgunday (dismissed):

Uzlopak (approved):

README.md
```diff
@@ -68,7 +68,7 @@ Prefix for the value of the cookie. This is useful for compatibility with `expre
 ##### cookie
 The options object is used to generate the `Set-Cookie` header of the session cookie. May have the following properties:
 * `path` - The `Path` attribute. Defaults to `/` (the root path).
-* `maxAge` - A `number` in milliseconds that specifies the `Expires` attribute by adding the specified milliseconds to the current date. If both `expires` and `maxAge` are set, then `maxAge` is used.
+* `maxAge` - A `number` in seconds that specifies the `Expires` attribute by adding the specified seconds to the current date. If both `expires` and `maxAge` are set, then `maxAge` is used.
```
jsumners, line 71: Can you point out where it is being interpreted as seconds? I understand the cookie spec defines it as seconds, but the implementation looks to be in milliseconds:

https://github.com/fastify/session/blob/25b4f6ba4814385a31293b13ddc5f57cac0cf8e2/lib/cookie.js#L42-L46

jsumners (changes requested):

README.md
likerRr: What are use cases setting `maxAge` in milliseconds? Seems like it's a common practice to set in seconds

gurgunday: Yeah, I think we should swap out this cookie module with the one from fastify/cookie

likerRr: But that would means that major version should be increased, which doesn't seem like a significant change) So probably a fix to the typings would be more then enough

gurgunday: I see - can you point to where the issue is in types?

jsumners: > What are use cases setting `maxAge` in milliseconds?

JavaScript's date handling is based upon milliseconds, not seconds.

chuanqisun:
Hey @gurgunday (or whom it may concern), sorry for bringing up a stale issue again. I believe there is still an issue here. This is my most recent observation.

In the typing file, `@fastify/session :: CookieOptions`  is already based on `@fastify/cookie :: CookieSerializeOptions` which interprets the value as seconds

[@fastify/session/types/types.d.ts](https://github.com/fastify/session/blob/8554dca6c74182ab753cb8f71347f2c97a869fb6/types/types.d.ts#L177)
```typescript
import { CookieSerializeOptions } from "@fastify/cookie"
...
export interface CookieOptions extends Omit<CookieSerializeOptions, 'signed'> {}
```

This means when user hover over the property, they will believe that the unit is seconds, which contradicts the implementation. 

Screenshot:
image

If we continue with the milliseconds implementation, more typescript users could build applications with the wrong assumption that the value is seconds. I see two options:

1. Fix the typing. Stop relying on @fastify/cookie's type. Something like this might work:
   ```typescript
   export interface CookieOptions extends Omit<CookieSerializeOptions, 'signed' | 'maxAge'> {
      /** A `number` in milliseconds that specifies the `Expires` attribute by adding the specified seconds to the current date. If both `expires` and `maxAge` are set, then `expires` is used. */
     maxAge?: number;
   }
   ```
2. Fix the implementation. Make it a breaking change. Because `@fastify/session` already uses `@fastify/cookies`, it seems natural to adopt the same unit.

gurgunday:
I think we should do both 😅

Can you first make a PR to main with the first alternative, and we can follow up with the second for the next branch

chuanqisun:
> I think we should do both 😅
> 
> Can you first make a PR to main with the first option, and we can follow up with the second for the next branch

See #245

gurgunday:
Thanks for the try!

PR #241, opened 2024-04-10
enhance checkOptions to reject invalid signer objects

autopulated:
I passed a Buffer instead of a string as a secret, and this resulted in an error message that "cookieSigner.sign is not a function" an at request time, rather than at plugin registration.

Someone else had a similar issue in the past https://github.com/fastify/help/issues/852

This PR makes `checkOptions` more comprehensive (explicitly checking for `.sign` and `.unsign` methods on signer objects passed as options.secret), and adds associated tests, so that this error is caught sooner and with a helpful message :)

#### Checklist

- [n/a] documentation is changed or added

### DIFF
--- a/lib/fastifySession.js
+++ b/lib/fastifySession.js
@@ -203,14 +203,16 @@ function fastifySession (fastify, options, next) {
   }
 
   function checkOptions (options) {
-    if (!options.secret) {
-      return new Error('the secret option is required!')
-    }
-    if (typeof options.secret === 'string' && options.secret.length < 32) {
-      return new Error('the secret must have length 32 or greater')
-    }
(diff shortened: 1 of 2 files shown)


autopulated:
@gurgunday It seems that the cookie "maxAge set by maxAge" test has a race condition, since the clock can run forwards 1ms between https://github.com/fastify/session/blob/master/test/cookie.test.js#L582 and https://github.com/fastify/session/blob/master/test/cookie.test.js#L584, changing the max age of the cookie from 1000 to 999.

gurgunday:
Yeah I remember now

gurgunday (approved):

mcollina (approved):
lgtm

PR #247, opened 2024-04-22
use sha256 instead of sha1 as default

mcollina:
#### Checklist

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -215,7 +215,7 @@ module.exports = class Session {
     })
 
     return crypto
-      .createHash('sha1')
+      .createHash('sha256')
       .update(str, 'utf8')
       .digest('hex')
   }


mcollina:
I'm not 100% sure if this is a semver-major change.

simoneb (approved):
Because this is not public API, I'd say it isn't a breaking change.

Uzlopak:
I am saying it is a semver major change.

Just one reason: sha256 is slower than sha1. If somebody upgrades this plugin and suddenly sees some perf drops...

Also i think we also did a major when we upgraded from sha1 to sha256 in a different plugin as default

mcollina:
@Uzlopak +1 as semver-major. You ok with the change?

SimenB:
Out of curiosity, why change it? It's only used for revving. The value only exists in memory, so any security issues with using it does not apply

mcollina:
False positive from security reviews, saves us time to have to justify it.

SimenB:
Hah, fair enough! 😀

mcollina:
@SimenB do you think this is a breaking change?

Uzlopak:
Ah. right. it is just for revving. But yeah, performance will foreseeable drop.

SimenB:
I wouldn't say so as it's opaque to the user. I don't think the performance degredation will be noticeable outside of micro benchmarks (but I might be wrong for a service serving thousands of requests per minute, which would result in lots of hash checks - I'd still be surprised if the hashing was noticable, tho). But maybe set up an autocannon run or something which triggers the hashing and see if it makes a difference?

SimenB (approved):

arthurfiorette:
Hello @mcollina @Uzlopak, I think this was a breaking change for us. We have more than 1 nodejs service running on different clients and/or versions and this subtle change broke almost all of our logins 😭 I have to admit, this was a pain to debug.

We pinned `10.7.2` on our side but I bet this will be happening to a lot more people than us, I'd recommend to revert this change and re-relase it under a new major version.

SimenB:
Huh, I didn't think the hash escaped memory - it shouldn't be persisted to storage and thus be affected by distributed systems. Unless I'm misremembering - it's been a while since this code was added 😅

simoneb:
> Huh, I didn't think the hash escaped memory - it shouldn't be persisted to storage and thus be affected by distributed systems. Unless I'm misremembering - it's been a while since this code was added 😅

I was trying to think about that too, because we all missed it. I guess the scenario is the session being stored in a cookie and rountripping across different instances of the server, running different versions of this package.

SimenB:
To be certain, you had issues with 10.8.0? Not 10.9.0 which has 0495ce5b534c4550f25228821db8098293439f2f (dunno if that would affect anything either, but that at least changes the data serialized to storage). Regardless, the hash itself is just to check "did the session data change since it was loaded from storage"

---

The hash shouldn't be in the cookie either (but I only ever stored the session id in the cookie, never the _values_. so in that case there might be edge cases I never considered)

PR #249, opened 2024-04-29
Fix type argument inference of `session.get(key)` and `session.set(key, value)`

zetaraku:
The current generic method `session.get(key)` accidentally allows the return value to be inferred into any type.

The `session.set(key, value)` also allows `value` of any type to be passed in.

Image

**Note:** This fix removes the second generic param `V`, so it will break an undocumented usage for TypeScript users specifying both two generic type parameters. (This should be rare.)

#### Checklist


mcollina (approved):
lgtm

Issue #257, opened 2024-06-30
Add the ability to configure cookie options within a route


ogheo:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Same as https://github.com/fastify/fastify-secure-session?tab=readme-ov-file#configuring-cookie-options-inside-a-route from fastify-secure-session.

### Motivation

In some cases, it might be really useful, like implementing a 'remember me' feature or something else, without big hassles. Continuation of https://github.com/fastify/session/pull/53.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

PR #260, opened 2024-07-04
Added the ability to configure cookie options within a route

ogheo:
This PR add the ability to configure cookie options within a route. Relates to https://github.com/fastify/session/issues/257

#### Checklist

- [X] run `npm run test` and `npm run benchmark`
- [X] tests and/or benchmarks are included
- [X] documentation is changed or added
- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/lib/session.js
+++ b/lib/session.js
@@ -56,6 +56,13 @@ module.exports = class Session {
     this[persistedHash] = this[hash]()
   }
 
+  options (opts) {
+    if (Object.keys(opts).length) {
+      this[cookieOptsKey] = opts
+      Object.assign(this.cookie, this[cookieOptsKey])
+    }
+  }
+
   touch () {
     if (this.cookie.originalMaxAge) {
       this.cookie.expires = new Date(Date.now() + this.cookie.originalMaxAge)
--- a/lib/session.js
+++ b/lib/session.js
@@ -58,8 +58,8 @@ module.exports = class Session {
 
   options (opts) {
     if (Object.keys(opts).length) {
-      this[cookieOptsKey] = opts
-      Object.assign(this.cookie, this[cookieOptsKey])
+      Object.assign(this[cookieOptsKey], opts)
+      this.cookie = new Cookie(this[cookieOptsKey], this[requestKey])
     }
   }
 
--- a/test/session.test.js
+++ b/test/session.test.js
@@ -1141,7 +1141,7 @@ test('Custom options', async t => {
 })
 
 test('Override global options', t => {
-  t.plan(7)
+  t.plan(11)
 
   const fastify = Fastify()
   fastify.register(fastifyCookie)
@@ -1156,20 +1156,22 @@ test('Override global options', t => {
 
   fastify.post('/', (request, reply) => {
     request.session.set('data', request.body)
-    request.session.options({ maxAge: 1000 * 60 * 60 })
+    request.session.options({ maxAge: 1000 * 60 * 60 }) // maxAge updated to 1 hour
 
     reply.send('hello world')
   })
 
   t.teardown(fastify.close.bind(fastify))
 
-  fastify.get('/', (request, reply) => {
+  fastify.get('/', async (request, reply) => {
     const data = request.session.get('data')
+    await request.session.regenerate()
 
     if (!data) {
       reply.code(404).send()
       return
     }
+
     reply.send(data)
   })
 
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

test/session.test.js
```diff
@@ -1083,3 +1083,119 @@ test('should save session if existing, modified, rolling false, and cookie.expir
...
+    cookie: {
+      secure: false,
+      maxAge: 42,
+      path: '/'
+    }
+  })
+
+  fastify.post('/', (request, reply) => {
+    request.session.set('data', request.body)
+    request.session.options({ maxAge: 1000 * 60 * 60 })
```
gurgunday, line 1159: Can you add one more route other than `/` (and a request to that route) that sets a different session data to show that `maxAge` is still 42 there?

lib/session.js
```diff
@@ -56,6 +56,13 @@ module.exports = class Session {
     this[persistedHash] = this[hash]()
   }
 
+  options (opts) {
+    if (Object.keys(opts).length) {
+      this[cookieOptsKey] = opts
+      Object.assign(this.cookie, this[cookieOptsKey])
```
gurgunday, line 62: So we are merging to `this.cookie` but directly assigning to `this[cookieOptsKey]`, aren't we losing its other properties with that direct assignment?

[`.regenerate`](https://github.com/fastify/session/pull/260/files#diff-1f2570998a9bf80486f544bb34c3bbf16e1af1755a37ac836eabbd82c6be4fb4R77-R83) will forget its `maxAge` if we modify `path` for instance, no?

Why not:

```suggestion
      Object.assign(this[cookieOptsKey], opts)
      this.cookie = this[cookieOptsKey]
```

ogheo: > So we are merging to this.cookie but directly assigning to this[cookieOptsKey], aren't we losing its other properties with that direct assignment?

I suppose you're right, better to merge old cookieOptsKey with newer ones, regarding setting direct to this.cookie I don't think it's a good idea since old params will be lost and we don't want to set maxAge but to use [cookie setter](https://github.com/fastify/session/blob/master/lib/cookie.js#L45), for example:

cookie before:
```
{
  path: '/',
  secure: true,
  sameSite: 'strict',
  domain: null,
  httpOnly: true,
  partitioned: undefined,
  _expires: null,
  originalMaxAge: null
}
```
cookie after:
```
{
  secure: true,
  httpOnly: true,
  sameSite: 'strict',
  maxAge: 2592000000
}
```

gurgunday: I created this test that fails currently (notice the usage of `regenerate`):

```js
test('#260', t => {
  t.plan(11)

  const fastify = Fastify()
  fastify.register(fastifyCookie)
  fastify.register(fastifySession, {
    ...DEFAULT_OPTIONS,
    cookie: {
      secure: false,
      maxAge: 42,
      path: '/'
    }
  })

  fastify.post('/', (request, reply) => {
    request.session.set('data', request.body)
    request.session.options({ maxAge: 1000 * 60 * 60 }) // maxAge updated to 1 hour

    reply.send('hello world')
  })

  t.teardown(fastify.close.bind(fastify))

  fastify.get('/', async (request, reply) => {
    const data = request.session.get('data')
    await request.session.regenerate()

    if (!data) {
      reply.code(404).send()
      return
    }

    reply.send(data)
  })

  fastify.inject({
    method: 'POST',
    url: '/',
    payload: {
      some: 'data'
    }
  }, (error, response) => {
    t.error(error)
    t.equal(response.statusCode, 200)
    t.ok(response.headers['set-cookie'])
    const { expires, path } = response.cookies[0]
    t.equal(expires.toUTCString(), new Date(Date.now() + 1000 * 60 * 60).toUTCString())
    t.equal(path, '/')

    fastify.inject({
      method: 'GET',
      url: '/',
      headers: {
        cookie: response.headers['set-cookie']
      }
    }, (error, response) => {
      t.error(error)
      t.equal(response.statusCode, 200)
      t.same(JSON.parse(response.payload), { some: 'data' })
      t.ok(response.headers['set-cookie'])
      const { expires, path } = response.cookies[0]
      t.equal(expires.toUTCString(), new Date(Date.now() + 1000 * 60 * 60).toUTCString()) // This will fail
      t.equal(path, '/')
    })
  })
})
```

We need this:

```suggestion
      Object.assign(this[cookieOptsKey], opts)
      this.cookie = new Cookie(this[cookieOptsKey], this[requestKey])
```

mcollina (approved):
lgtm

gurgunday (changes requested):
https://github.com/fastify/session/pull/260#discussion_r1670012257 needs resolution

gurgunday (commented):
Looks good to me

One last thing, can you also add this [test](https://github.com/fastify/session/pull/260#discussion_r1670012257) that tests the feature with `regenerate`?

Thanks

ogheo:
@gurgunday, I updated Override global options test with regenerate changes, is that ok or would you like to add it as a separate test?

gurgunday (approved):
lgtm

ogheo:
@mcollina @gurgunday can we merge or something else is required?

PR #271, opened 2024-10-10
use 'lax' instead of 'Lax' as default for cookie sameSite option when secure is set to 'auto'

gunjam:
Change to support @fastify/cookie ^11.0.0

When setting the secure option to 'auto' the cookie option of 'Lax' is being set by default for non HTTPS traffic, which is causing the jshttp/cookie module to throw as it now only supports the [lowercase option of 'lax'](https://github.com/jshttp/cookie/blob/master/src/index.ts#L332).

Had to update the tests to check for maxAge 0 on expiry which @fastify/cookie is now also [having to set](https://github.com/fastify/fastify-cookie/blame/master/plugin.js#L44) due to changes in jshttp/cookie.

### DIFF
--- a/lib/cookie.js
+++ b/lib/cookie.js
@@ -28,7 +28,7 @@ module.exports = class Cookie {
       if (request.protocol === 'https') {
         this.secure = true
       } else {
-        this.sameSite = 'Lax'
+        this.sameSite = 'lax'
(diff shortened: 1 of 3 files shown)


gurgunday (approved):
lgtm

PR #293, opened 2025-01-23
Implemented way to customize session-id retrieval/storage

Levyks:
#### Checklist


lib/idStore.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
mcollina, line 1:
```suggestion
'use strict'

module.exports = {
```

mcollina (changes requested):
Can you add a test for the types? We use tsd

PR #309, opened 2025-08-07
localhost is also a secure origin by browsers

segevfiner:
Also partitioned cookies are only allowed with secure, so in `secure: 'auto'` also set partitioned to false

#### Checklist

- [ ] run `npm run test && npm run benchmark --if-present`
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/test/cookie.test.js
+++ b/test/cookie.test.js
@@ -318,6 +318,7 @@ test('should set session cookie secureAuto', async (t) => {
   t.after(() => { fastify.close() })
 
   const response = await fastify.inject({
+    authority: 'fastify-session.test',
     url: '/'
   })
 
@@ -345,6 +346,7 @@ test('should set session cookie secureAuto change SameSite', async (t) => {
   t.after(() => { fastify.close() })
 
   const response = await fastify.inject({
+    authority: 'fastify-session.test',
     url: '/'
   })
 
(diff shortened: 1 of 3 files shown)


Eomm (changes requested):
Marking to review it

segevfiner (commented):
Tried to fix that tests. Seems there was an implicit assumption about `localhost` being insecure in some of the tests, so I tried to work around that.

Eomm (approved):

mcollina (approved):
lgtm

test/cookie.test.js
```diff
@@ -314,6 +314,7 @@ test('should use session cookie secure override when saving secure cookie', asyn
   t.after(() => { fastify.close() })
 
   const response = await fastify.inject({
+    authority: 'fastify-session.test',
```
mcollina, line 317: why is this now needed

mcollina (changes requested):

test/cookie.test.js
segevfiner: Same as the other tests. It defaulted to localhost, but those are tests that explictly require a domain.

mcollina: How come? These are local and done via .inject()

segevfiner: Yes. And as far as I remember when I debugged it. Inject defaulted to localhost as the authority, while the tests expected an insecure request, as the previously expected localhost to be insecure.

PR #324, opened 2026-03-08
fix: isolate session cookie option overrides

mcollina:
## Summary
- isolate cookie option state per session by cloning cookie options in `Session`
- keep per-session `session.options()` overrides local to that session instead of mutating shared plugin options
- preserve override behavior across reload/regenerate without leaking stale/expired absolute cookie expiration values
- add regression tests for cross-session isolation and missing `maxAge` getter branch coverage

## Context
This addresses a real bug where `session.options()` could mutate shared cookie option state. A route still needs to pass untrusted input into `session.options()` for security impact, so this fix is framed as state isolation correctness and hardening.

## Testing
- `npm test`
- `npm run test:typescript`


gurgunday (approved):
lgtm

PR #342, opened 2026-07-12
fix session secure cookie override

0XFF-96:
## Summary 
Fix secure cookie override handling so the insecure-connection guard uses the effective per-session cookie config instead of the global default.

## Details

`Session#options()` can override cookie options for the current session. 
Previously, `onSend` checked `options.cookie.secure`, while the emitted Set-Cookie used `session.cookie.toJSON()`. 

This caused mismatched behavior when secure was overridden per request.

This change makes the guard check `session.cookie.secure`, matching the actual cookie that will be saved/sent.

#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/index.js
+++ b/index.js
@@ -166,7 +166,7 @@ function fastifySession (fastify, options, next) {
 
       const cookieSessionId = getCookieSessionId(request)
       const saveSession = shouldSaveSession(request, cookieSessionId, saveUninitializedSession, rollingSessions)
-      const isInsecureConnection = cookieOpts.secure === true && request.protocol !== 'https'
+      const isInsecureConnection = session.cookie.secure === true && request.protocol !== 'https'
       const sessionIdWithPrefix = hasCookiePrefix ? `${cookiePrefix}${session.encryptedSessionId}` : session.encryptedSessionId
       if (!saveSession || isInsecureConnection) {
         // if a session cookie is set, but has a different ID, clear it
--- a/test/base.test.js
+++ b/test/base.test.js
@@ -3,7 +3,7 @@
 const test = require('node:test')
 const Signer = require('@fastify/cookie').Signer
 const fastifyPlugin = require('fastify-plugin')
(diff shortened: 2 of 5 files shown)


0XFF-96:
## Context

Repo docs say cookie options are used to generate the session Set-Cookie header: [README.md (line 68)](https://github.com/fastify/session/blob/main/README.md). They also document Session#options(opts) as a route-level way to update cookie options: [README.md (line 125)](https://github.com/fastify/session/blob/main/README.md).


So the intended chain is:

```text
global cookie options = default
Session#options() = per-request override
session.cookie = final effective cookie
```

`Session#options()` confirms this in code by merging the override into current session cookie options and rebuilding session.cookie: [lib/session.js (line 82)]().


## Before vs After
Before this PR, the insecure-connection guard checked the global default:

```text
cookieOpts.secure === true && request.protocol !== 'https'
But the actual cookie sent to the client used the effective session cookie:
session.cookie.toJSON()
```
That created a mismatch: decision from global config, output from per-session config.

After this PR, the guard checks the same effective cookie that will be emitted:

```text
session.cookie.secure === true && request.protocol !== 'https'
```

## Reference 

[RFC 6265 section 4.1.2.5](https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.5) defines Secure: browsers should only send that cookie over a secure channel, typically HTTPS over TLS.
[OWASP Session Management Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html) recommends session cookies use Secure to avoid exposing session IDs over unencrypted traffic.

mcollina (approved):
lgtm

Tony133 (approved):

PR #343, opened 2026-07-12
fix: use named import for connect-redis v8.x

Tony133:
Proposal:

- Bumps `connect-redis` to `^8.0.0` and updates the imports that broke because of it ( see v8.0.0 here: https://github.com/tj/connect-redis/releases/tag/v9.0.0)

Changes:

- Since v8 ( see v8.0.0 here: https://github.com/tj/connect-redis/releases/tag/v8.0.0), `connect-redis` no longer has a default export `RedisStore` is now a named export only. This broke:

  - `types/index.tst.ts` — type-check failure (`tstyche`, TS2351: "This expression is not constructable")
  - `examples/redis.js` and `benchmark/bench.js` — runtime error, since`require('connect-redis').default` resolves to `undefined`

### DIFF
--- a/benchmark/bench.js
+++ b/benchmark/bench.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const RedisStore = require('connect-redis').default
+const { RedisStore } = require('connect-redis')
 const Fastify = require('fastify')
 const Redis = require('ioredis')
 const fileStoreFactory = require('session-file-store')
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "@types/node": "^26.0.1",
     "c8": "^11.0.0",
     "connect-mongo": "^6.0.0",
-    "connect-redis": "^7.1.1",
+    "connect-redis": "^8.0.0",
     "cronometro": "^6.0.3",
     "eslint": "^9.17.0",
     "fastify": "^5.0.0",
--- a/examples/redis.js
+++ b/examples/redis.js
@@ -4,7 +4,7 @@ const Fastify = require('fastify')
 const fastifySession = require('..')
 const fastifyCookie = require('@fastify/cookie')
 const Redis = require('ioredis')
-const RedisStore = require('connect-redis').default
+const { RedisStore } = require('connect-redis')
 
 const fastify = Fastify()
 
--- a/types/index.tst.ts
+++ b/types/index.tst.ts
@@ -1,5 +1,5 @@
 import MongoStore from 'connect-mongo'
-import RedisStore from 'connect-redis'
+import { RedisStore } from 'connect-redis'
 import fastify, {
   type FastifyInstance,
   type FastifyReply,


package.json
```diff
@@ -65,7 +65,7 @@
     "@types/node": "^26.0.1",
     "c8": "^11.0.0",
     "connect-mongo": "^6.0.0",
-    "connect-redis": "^7.1.1",
+    "connect-redis": "^9.0.0",
```
is2ei: Since v9.0.0, `connect-redis` package no longer supports `ioredis`.

https://github.com/tj/connect-redis/releases
> This library no longer supports ioredis. It only supports redis (v5 and later) now.

Should we also update the example to use `redis` instead of `ioredis`?

https://github.com/fastify/session/blob/fix/connect-redis-named-export-types/examples/redis.js#L6

Tony133: > Since v9.0.0, `connect-redis` package no longer supports `ioredis`.
> 
> https://github.com/tj/connect-redis/releases
> 
> > This library no longer supports ioredis. It only supports redis (v5 and later) now.
> 
> Should we also update the example to use `redis` instead of `ioredis`?
> 
> https://github.com/fastify/session/blob/fix/connect-redis-named-export-types/examples/redis.js#L6

Maybe instead of version 9, I'll set the dependency to version 8, since it was still compatible for now; then we'll figure out which Redis packages to use.

Tony133: Done, I've updated to version v8.x., see commit here: [3363308](https://github.com/fastify/session/pull/343/commits/336330892a2b1dd506bffe53f405f849affe7683)

benchmark/bench.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const RedisStore = require('connect-redis').default
+const { RedisStore } = require('connect-redis').default
```
is2ei: @Tony133 
The code example in the official migration guide does not use `default`:
https://github.com/tj/connect-redis/releases/tag/v8.0.0

Image

Could you confirm that this change is correct?

package.json
is2ei: @Tony133 
Thanks!
How about explicitly updating it to [v8.1.0](https://github.com/tj/connect-redis/releases/tag/v8.1.0)? It includes a few bug fixes.

Tony133: > @Tony133 Thanks! How about explicitly updating it to [v8.1.0](https://github.com/tj/connect-redis/releases/tag/v8.1.0)? It includes a few bug fixes.

actually download `v8.1.0` already if you see in `node_modules`

benchmark/bench.js
Tony133: > @Tony133 The code example in the official migration guide does not use `default`: https://github.com/tj/connect-redis/releases/tag/v8.0.0
> 
> Image
> Could you confirm that this change is correct?

correct, see commit here: [6d0a844](https://github.com/fastify/session/pull/343/commits/6d0a8449a9b4833ab418fc684d930e1a637c502c)

is2ei: @Tony133 

Could you also confirm this line?

Tony133: > @Tony133
> 
> Could you also confirm this line?


done! see commit here: [9a8c488](https://github.com/fastify/session/pull/343/commits/9a8c48857d7570fb1dfb0587065885673ca3f4c8), I also added a script in the `package.json` that launches a Docker container with redis, making both the example and the benchmark more convenient to run ( see here: [bf571cb](https://github.com/fastify/session/pull/343/commits/bf571cb4138c52eb0dbe99e572124571c8310ab8))

is2ei: @Tony133 

> I also added a script in the package.json

The change isn't about updating the `connect-redis` package version, so I don't think we should include it in this PR.

Tony133: > @Tony133
> 
> > I also added a script in the package.json
> 
> The change isn't about updating the `connect-redis` package version, so I don't think we should include it in this PR.

Okay, I've removed it; we'll put it in another PR.

is2ei (approved):

PR #348, opened 2026-07-16
fix: destroy the previous session on regenerate

### LINKED ISSUE #240: `Session#regenerate` does not call `SessionStore#destroy`
### Prerequisites

### Fastify version

4.26.2

### Plugin version

10.7.0

### Node.js version

20.11.1

### Operating system

Linux

5.15.150-1-MANJARO

### Description

I am trying to regenerate a session in one of my routes to make sure whatever data is in there is reset and I would prefer to also get a new session id as I am essentially restarting the session.

But as far as I can see there is no way for me, without manually interacting with the *SessionStore* myself, to delete the old session id.

`#regenerate` only generates a new session for me and stores it.

### Steps to Reproduce

Something like:
```ts
this.fastify.get('/regenerate', async (request, reply) => {
  await request.session.regenerate();
  request.session.set('userId', 123);
  await request.session.save();

  return reply.send();
});
```

For easier understanding on what is happening in the store:
```ts
    this.fastify.register(FastifySession, {
      secret: '...',
      store: {
        set: (sessionId, session, callback) => {
          console.log(`#set(${JSON.stringify(sessionId)}, ${JSON.stringify(session)}`);
          callback();
        },
        get: (sessionId, callback) => {
          console.log(`#get(${JSON.stringify(sessionId)})`);
          callback(null, null);
        },
        destroy: (sessionId, callback) => {
          console.log(`#destroy(${JSON.stringify(sessionId)})`);
          callback();
        }
      }
    });
```

### Expected Behavior

I am essentially generating a completely new session with different ID and data/content.

I'd expect the session to be automatically deleted from the store as it is no longer used/needed.

As far as I can see, there is also no good workaround for this as calling `await request.session.destroy();` before regenerating the session sets it to `null` causing `TypeError: Cannot read properties of null (reading 'regenerate')`

xianjianlf2:
`Session#regenerate` creates and stores a new session but never removes the previous one from the store, leaving a stale entry behind (issue #240). Because the old session id stays valid in the store, this can contribute to session fixation, as noted in the issue thread.

This destroys the old session id before storing the regenerated one, mirroring `express-session`'s `store.regenerate` behavior (destroy old, then issue new). Any `destroy` error is surfaced through the same callback/promise as before (`destroyError || setError`).

The expired-session branch in the `onRequest` hook already called `destroy` explicitly before `regenerate`; that is now redundant and has been removed, so an expired session results in a single `destroy` rather than two.

Added a test asserting that after `regenerate` the previous session id is no longer present in the store while the new one is. Full suite passes at 100% coverage.

Closes #240

### DIFF
--- a/index.js
+++ b/index.js
@@ -95,14 +95,8 @@ function fastifySession (fastify, options, next) {
       const expiration = restoredSession.cookie.originalExpires || restoredSession.cookie.expires
 
       if (expiration && expiration.getTime() <= Date.now()) {
-        restoredSession.destroy(err => {
-          if (err) {
-            done(err)
-            return
-          }
-
-          restoredSession.regenerate(done)
-        })
+        // `regenerate` destroys the expired session before issuing a new one.
+        restoredSession.regenerate(done)
         return
       }
 
--- a/lib/session.js
+++ b/lib/session.js
@@ -111,22 +111,29 @@ module.exports = class Session {
       }
     }
 
-    if (callback) {
-      this[sessionStoreKey].set(session.sessionId, session, error => {
-        this[requestKey].session = session
+    const oldSessionId = this[sessionIdKey]
 
-        callback(error)
+    if (callback) {
+      this[sessionStoreKey].destroy(oldSessionId, destroyError => {
+        this[sessionStoreKey].set(session.sessionId, session, setError => {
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

